PR #13, opened 2024-01-19
Created Launcher subsystem

iamawesomecat:
added multiple constant variables for shoot speed, added 3 motors and set portID constants.


src/main/java/frc/team2412/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,61 @@
...
+
+public class ShooterSubsystem extends SubsystemBase {
+	// CONSTANTS
+	// MOTOR SPEED VALUES
+	public static final double SPEAKER_SHOOT_SPEED = 0.8;
+	public static final double AMP_SHOOT_SPEED = 0.2;
+	public static final double ANGLE_CHANGE_SPEED = 0.15;
+
+	// HARDWARE
+	private final TalonFX shooterTopMotor;
```
kirbt: I would take away the final modifier from these hardware objects. Having final here will make it so that you are unable to modify the variables and thus even though it doesn't show any errors when you attempt to, you will not be able to actually set the speed of the motors, etc.

kirbt: Its good to use final on constants though, since they aren't supposed to be modified or changed

TAKBS2412: Final only means that the variable can't be assigned to another value, so operations like setting the speed of the motors will still work. ([Here's an example](https://github.com/robototes/ChargedUp2023/blob/main/src/main/java/frc/team2412/robot/subsystems/ArmSubsystem.java#L163-L165) from last year where we used final on motor controller objects).

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,87 @@
...
+					.withWidget(BuiltInWidgets.kTextView)
+					.getEntry();
+	// Constructor
+	public LauncherSubsystem() {
+
+		// MOTOR INSTANCE VARIBLES
+		launcherTopMotor = new CANSparkFlex(Hardware.LAUNCHER_TOP_MOTOR_ID, MotorType.kBrushless);
+		launcherBottomMotor = new CANSparkFlex(Hardware.LAUNCHER_BOTTOM_MOTOR_ID, MotorType.kBrushless);
+		launcherAngleMotor = new CANSparkFlex(Hardware.LAUNCHER_ANGLE_MOTOR_ID, MotorType.kBrushless);
+		launcherHoodMotor = new CANSparkFlex(Hardware.LAUNCHER_HOOD_MOTOR_ID, MotorType.kBrushless);
```
jamesdooley4: Need to configure the motors. See https://github.com/robototes/ChargedUp2023/blob/f25191718390d11df9ddda789a0f0d5c35686734/src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java#L124 for an example how

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	// stop specific motor method
+	public void stopMotor(CANSparkFlex motor) {
+		motor.stopMotor();
+	}
+
+	public void shoot(double speed) {
+		launcherTopMotor.set(speed);
+		launcherBottomMotor.set(-speed);
```
jamesdooley4: Better to invert the bottom motor & set both to the same target speed

jamesdooley4: Even better to not set a raw power (which results in different velocities in each motor) and instead get the motor's PID controller and call setReference(value, ControlType.kVelocity) to tell the motors to run at the same velocity

jamesdooley4 (changes requested):

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,87 @@
...
+	public double getHoodAngle() {
+		return Units.rotationsToDegrees(launcherHoodEncoder.getPosition());
+	}
+
+	public void setHoodAngle(double angle) {
+		launcherHoodPIDController.setReference(Units.degreesToRotations(angle), ControlType.kPosition);
+	}
+
+	public void DistanceModifier(double distance, double modifier) {
+		double speed = distance / modifier;
```
TAKBS2412: Unfinished?

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,87 @@
...
+	private final CANSparkFlex launcherTopMotor;
+	private final CANSparkFlex launcherBottomMotor;
+	private final CANSparkFlex launcherAngleMotor;
+	private final CANSparkFlex launcherHoodMotor;
+	private final SparkAbsoluteEncoder launcherAngleEncoder;
+	private final SparkAbsoluteEncoder launcherHoodEncoder;
+	private final SparkPIDController launcherAnglePIDController;
+	private final SparkPIDController launcherHoodPIDController;
+
+	GenericEntry launcherSpeed =
```
TAKBS2412: Not used - was the plan to add logging later?

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: i did that today, thanks!

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: Ill look into doing this, as for PID for the shooter ill get into that some time this week.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: removed for now, ty

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: fixed (i think, you can add a new issue if im wrong)

src/main/java/frc/team2412/robot/Hardware.java
```diff
@@ -9,8 +9,14 @@ public class Hardware {
...
+	// launcher [30 - 39]
+	public static final int LAUNCHER_TOP_MOTOR_ID = 30;
+	public static final int LAUNCHER_BOTTOM_MOTOR_ID = 31;
+	public static final int LAUNCHER_ANGLE_MOTOR_ID = 32;
+	public static final int LAUNCHER_HOOD_MOTOR_ID = 33;
 	// intake [40 - 49]
 
+	// sensors [1 - 19]
+	public static final int LAUNCHER_SENSOR_1 = 1;
+	public static final int LAUNCHER_SENSOR_2 = 2;
```
jamesdooley4: It'd be better to name these based on where they are on the robot

src/main/java/frc/team2412/robot/Subsystems.java
```diff
@@ -12,17 +13,21 @@ public static class SubsystemConstants {
...
 		public static final boolean LIMELIGHT_ENABLED = false;
 		public static final boolean CLIMB_ENABLED = false;
-		public static final boolean SHOOTER_ENABLED = false;
+		public static final boolean LAUNCHER_ENABLED = false;
 		public static final boolean INTAKE_ENABLED = false;
 		public static final boolean DRIVEBASE_ENABLED = true;
 	}
 
 	public final DrivebaseSubsystem drivebaseSubsystem;
+	public LauncherSubsystem launcherSubsystem;
```
jamesdooley4: Should be `final`

src/main/java/frc/team2412/robot/commands/launcher/LauncherPreSetAngle.java
```diff
@@ -0,0 +1,27 @@
+package frc.team2412.robot.commands.launcher;
+
+import edu.wpi.first.math.MathUtil;
+import edu.wpi.first.wpilibj2.command.Command;
+import frc.team2412.robot.subsystems.LauncherSubsystem;
+
+public class LauncherPreSetAngle extends Command {
```
jamesdooley4: This is named like a thing/noun instead of a action verb, which is confusing. Consider naming this `SetLauncherAngle` instead.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,140 @@
...
+import edu.wpi.first.wpilibj.DigitalInput;
+import edu.wpi.first.wpilibj.shuffleboard.BuiltInWidgets;
+import edu.wpi.first.wpilibj.shuffleboard.Shuffleboard;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.team2412.robot.Hardware;
+
+public class LauncherSubsystem extends SubsystemBase {
+	// CONSTANTS
+	// MOTOR SPEED VALUES
+	public static final double SPEAKER_SHOOT_SPEED = 0.5;
```
jamesdooley4, line 22: These seem like power levels (% of current battery voltage, which varies over a match) and not velocity values (which are the same regardless of battery voltage). It'd be more consistent to use velocity units here.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	public void configMotors() {
+		launcherTopMotor.restoreFactoryDefaults();
+		launcherBottomMotor.restoreFactoryDefaults();
+		launcherAngleMotor.restoreFactoryDefaults();
+		launcherHoodMotor.restoreFactoryDefaults();
+
+		// idle mode (wow)
+		launcherTopMotor.setIdleMode(IdleMode.kBrake);
```
jamesdooley4: The flywheel motors should be set to coast mode so they smoothly spin down when power is set to 0 vs. doing an abrupt stop (which uses a lot of energy to brake the flywheel).

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,140 @@
...
+		launcherBottomMotor.setIdleMode(IdleMode.kBrake);
+		launcherAngleMotor.setIdleMode(IdleMode.kBrake);
+		launcherHoodMotor.setIdleMode(IdleMode.kBrake);
+
+		// current limit
+		launcherTopMotor.setSmartCurrentLimit(20);
+		launcherBottomMotor.setSmartCurrentLimit(20);
```
jamesdooley4, line 87: Likely should set the bottom motor to run inverted from the top motor, and also have it follow the top motor's settings so they run at the same velocity. Though disregard if there's a need to run the top & bottom motors at different velocities.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
jamesdooley4: Then for now, just invert the bottom motor & set it up as a follower of the top motor, so this code only needs to set the top motor power & the bottom motor will automatically follow.

jamesdooley4: And file an issue to track the follow-up work to change from power units to velocity units with PIDF

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,140 @@
...
+		return Units.rotationsToDegrees(launcherHoodEncoder.getPosition());
+	}
+
+	public void setHoodAngle(double angle) {
+		launcherHoodPIDController.setReference(Units.degreesToRotations(angle), ControlType.kPosition);
+	}
+
+	@Override
+	public void periodic() {
+		launcherSpeed.setDouble(launcherTopEncoder.getVelocity());
```
jamesdooley4: This is reporting in different units (velocity) than the motor's being set at (raw power). It'd be better to be consistent and use power units here for now.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,140 @@
...
+	// CONSTANTS
+	// MOTOR SPEED VALUES
+	public static final double SPEAKER_SHOOT_SPEED = 0.5;
+	public static final double AMP_SHOOT_SPEED = 0.2;
+	public static final double ANGLE_CHANGE_SPEED = 0.15;
+	public static final double ANGLE_TOLERANCE = 0.5;
+
+	// HARDWARE
+	private final CANSparkFlex launcherTopMotor;
+	private final CANSparkFlex launcherBottomMotor;
```
jamesdooley4, line 29: To match the current prototype & likely robot launcher, I'd recommend also adding at least one (or two) motors for the feed rollers. Consider adding a `feedTopMotor` and `feedBottomMotor` with associated methods to set their power & report it to the dashboard.

jamesdooley4 (changes requested):

src/main/java/frc/team2412/robot/Hardware.java
iamawesomecat: Yeah, we will change it once we have a idea on every sensor on the bot and number it accordingly, ty!

src/main/java/frc/team2412/robot/Subsystems.java
iamawesomecat: done, ty!

src/main/java/frc/team2412/robot/commands/launcher/LauncherPreSetAngle.java
iamawesomecat: Updated, ty!

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: we will do this once we establish PID for the flywheels, but ty for letting me know!

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: done, thanks!

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: There will be a need to run them at different velocities because of the way we want to score AMP unfortunately

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: alright, set it to getVelocity for now, also added a comment to remind us to fix it once we establish that.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: talked about during meeting

src/main/java/frc/team2412/robot/commands/launcher/LauncherPreSetAngle.java
jamesdooley4: ? The class name hasn't changed...

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
jamesdooley4: FYI, in a later conversation with Marcus, there are two halves of index - one attached to the intake & one attached to the launcher. But that part of the robot was also going under a lot of design changes, so it's fine for now to leave those motors out from here.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,133 @@
...
+		launcherTopMotor = new CANSparkFlex(Hardware.LAUNCHER_TOP_MOTOR_ID, MotorType.kBrushless);
+		launcherBottomMotor = new CANSparkFlex(Hardware.LAUNCHER_BOTTOM_MOTOR_ID, MotorType.kBrushless);
+		launcherAngleMotor = new CANSparkFlex(Hardware.LAUNCHER_ANGLE_MOTOR_ID, MotorType.kBrushless);
+		launcherHoodMotor = new CANSparkFlex(Hardware.LAUNCHER_HOOD_MOTOR_ID, MotorType.kBrushless);
+		// encoders
+		launcherTopEncoder = launcherTopMotor.getEncoder();
+		launcherBottomEncoder = launcherBottomMotor.getEncoder();
+		launcherAngleEncoder = launcherAngleMotor.getAbsoluteEncoder(Type.kDutyCycle);
+		launcherHoodEncoder =
+				launcherHoodMotor.getAbsoluteEncoder(SparkAbsoluteEncoder.Type.kDutyCycle);
```
jamesdooley4: Should be able to remove `SparkAbsoluteEncoder.` here

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,133 @@
...
+		launcherTopMotor.burnFlash();
+		launcherBottomMotor.burnFlash();
+		launcherAngleMotor.burnFlash();
+		launcherHoodMotor.burnFlash();
+	}
+
+	// stop specific motor method
+	public void stopMotor(CANSparkFlex motor) {
+		motor.stopMotor();
+	}
```
jamesdooley4: Shouldn't be public, as nothing outside this class should be passing a motor as a parameter.  Since it's never used, consider deleting it, or replacing with something like `stopLauncher()` that sets the two launcher motors to 0 power.

jamesdooley4 (changes requested):

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
TAKBS2412: Agreed - though if we do keep it, we should make sure the motors aren't trying to fully stop (which could cause physical damage to the system) and are instead not contributing any more kinetic energy to the system (thinking back to 2022). If we do go down that route `stopMotor()` should work fine provided the other refactoring advice is followed.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,133 @@
...
+	private final CANSparkFlex launcherAngleMotor;
+	private final CANSparkFlex launcherHoodMotor;
+	private final RelativeEncoder launcherTopEncoder;
+	private final RelativeEncoder launcherBottomEncoder;
+	private final SparkAbsoluteEncoder launcherAngleEncoder;
+	private final SparkAbsoluteEncoder launcherHoodEncoder;
+	private final SparkPIDController launcherAnglePIDController;
+	private final SparkPIDController launcherHoodPIDController;
+
+	GenericEntry launcherSpeed =
```
TAKBS2412: Nit: this variable and the other `GenericEntry` should be `private final`

TAKBS2412 (changes requested):

src/main/java/frc/team2412/robot/commands/launcher/LauncherPreSetAngle.java
iamawesomecat: oops, fixed it now

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: i gotchu

jamesdooley4 (dismissed):

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: i went with the stopLauncher idea, ty!

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	public void setAngle(double angle) {
+		launcherAnglePIDController.setReference(Units.degreesToRotations(angle), ControlType.kPosition);
+	}
+
+	@Override
+	public void periodic() {
+		// .get will be replaced with .getVelocity once PID is established for flywheels :C
+		launcherSpeedEntry.setDouble(launcherTopMotor.get());
```
jamesdooley4: When the launcher stops, I think this will set the launcher setpoint used by `shoot()` to zero, which is unexpected. It'd be better to have separate dashboard entries for the current speed/power, and the desired speed/power used when `shoot()` is called.

jamesdooley4 (changes requested):

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: How would i get the current speed then?

iamawesomecat: i think i figured this out, ty!

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,121 @@
...
+
+	// stop specific motor method
+	public void stopLauncher() {
+		launcherTopMotor.stopMotor();
+		launcherBottomMotor.stopMotor();
+	}
+
+	public void launch() {
+		launcherTopMotor.set(launcherSpeedEntry.getDouble(SPEAKER_SHOOT_SPEED));
+		launcherBottomMotor.set(launcherSpeedEntry.getDouble(SPEAKER_SHOOT_SPEED));
```
KangarooKoala:
```suggestion
        double speed = launcherSpeedEntry.getDouble(SPEAKER_SHOOT_SPEED);
		launcherTopMotor.set(speed);
		launcherBottomMotor.set(speed);
```
This guarantees that the the motors are set to the same speed (in case the entry gets set to a different value between the calls).

jamesdooley4 (dismissed):

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,121 @@
...
+					.withProperties(Map.of("Min", -1, "Max", 1))
+					.getEntry();
+
+	private final GenericEntry launcherAngleEntry =
+			Shuffleboard.getTab("Launcher")
+					.addPersistent("Launcher angle", getAngle())
+					.withSize(1, 1)
+					.withWidget(BuiltInWidgets.kTextView)
+					.getEntry();
+	private final GenericEntry launcherSpeed =
```
KangarooKoala: Please pick more descriptive names for this and `launcherSpeedEntry` to clarify that one is set by the user and the other is set by robot code.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,121 @@
...
+			Shuffleboard.getTab("Launcher")
+					.addPersistent("Launcher Speed setpoint", SPEAKER_SHOOT_SPEED)
+					.withSize(2, 1)
+					.withWidget(BuiltInWidgets.kNumberSlider)
+					.withProperties(Map.of("Min", -1, "Max", 1))
+					.getEntry();
+
+	private final GenericEntry launcherAngleEntry =
+			Shuffleboard.getTab("Launcher")
+					.addPersistent("Launcher angle", getAngle())
```
KangarooKoala: Should this and `launcherSpeed` be persistent? I don't think we want the value to be saved across runs of the robot, and I could imagine this causing confusion by hiding when the robot isn't connected.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -0,0 +1,121 @@
...
+		launcherBottomMotor.setSmartCurrentLimit(20);
+		launcherAngleMotor.setSmartCurrentLimit(20);
+
+		launcherTopMotor.burnFlash();
+		launcherBottomMotor.burnFlash();
+		launcherAngleMotor.burnFlash();
+	}
+
+	// stop specific motor method
+	public void stopLauncher() {
```
KangarooKoala: Please either update the comment or delete it.

src/main/java/frc/team2412/robot/commands/launcher/LaunchCommand.java
KangarooKoala, line 1: As a note, because this command only sets the motor speeds when it (the command) is initialized, changes to the launcher speed setpoint entry will not be updated until the command is scheduled again.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: Okay, i did exactly that, tysm!

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: Fixed, ty

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: no it shouldnt lol, thats my bad. fixed now tho :D

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: done, ty

TAKBS2412 (dismissed):
LGTM pending Joseph's comments (note the PR will have to be re-approved as adding new commits removes existing approvals).

src/main/java/frc/team2412/robot/commands/launcher/LaunchCommand.java
iamawesomecat: fixed, ty!

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -33,7 +33,7 @@ public class LauncherSubsystem extends SubsystemBase {
 	private final SparkAbsoluteEncoder launcherAngleEncoder;
 	private final SparkPIDController launcherAnglePIDController;
 
-	private final GenericEntry launcherSpeedEntry =
+	private final GenericEntry SetlauncherSpeedEntry =
```
jamesdooley4: Probably should be lowercase to start, to match the other class variables: `setLauncherSpeedEntry`

jamesdooley4 (dismissed):

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -103,6 +103,11 @@ public void launch() {
 		launcherTopMotor.set(speed);
 		launcherBottomMotor.set(speed);
```
TAKBS2412: (this could be renamed and rewritten to use the other `launch()` method... but I think these methods will change anyway once we switch to velocity instead of power, so probably fine to leave for now).

TAKBS2412 (dismissed):

jamesdooley4 (approved):

TAKBS2412 (approved):

PR #15, opened 2024-01-23
Created Climb Subsystem

Yumnah2:
Created Climb Subsystem

### DIFF
--- a/src/main/java/frc/team2412/robot/subsystems/ClimbSubsystem.java
+++ b/src/main/java/frc/team2412/robot/subsystems/ClimbSubsystem.java
@@ -5,40 +5,48 @@ import edu.wpi.first.wpilibj2.command.Command;
 import frc.team2412.robot.Robot;
 
 import com.revrobotics.SparkMaxAbsoluteEncoder;
-import com.revrobotics.CANSparkMax;
-import com.revrobotics.CANSparkMaxLowLevel.MotorType;
-import com.revrobotics.CANSparkMax.IdleMode;
+import com.revrobotics.CANSparkFlex;
+import com.revrobotics.CANSparkFlexLowLevel.MotorType;
+import com.revrobotics.CANSparkFlex.IdleMode;
 
 public class ClimbSubsystem extends SubsystemBase{
     //public or private?
-    //  CONSTANT VARIABLES
+    //CONSTANT VARIABLES
         //speeds (fix later)
     private final double EXTEND_SPEED = 0.1;
     private final double RETRACT_SPEED = -0.1;
-    private final double MOTOR_SPEED = 0.15;
+    //private final double MOTOR_SPEED = 0.15;
         //PID extension values
         //PID retraction values
         //height of chain in encoder??? units
 
-    //  INSTANCE VARIABLES 
-    private final CANSparkMax motor1;
-    private final CANSparkMax motor2;
-    private final SparkMaxAbsoluteEncoder encoder;
+    //  INSTANCE VARIABLES
+    private final CANSparkFlex motorLeft;
+    private final CANSparkFlex motorRight;
+    private final SparkFlexAbsoluteEncoder encoderLeft;
+    private final SparkFlexAbsoluteEncoder encoderRight;
+    private final SparkPIDController motorLeftPIDController;
+    private final SparkPIDController motorRightPIDController;
 
     //encoder position
     public ClimbSubsystem() {
         //motor = new MOTOR_NAME (port number found in hardware);
-        this.motor1 = new motor(CLIMB_MOTOR_ONE, MotorType.kBrushless);
-        this.motor2= new motor(CLIMB_MOTOR_TWO,MotorType.kBrushless );
-        this.encoder1 = new motor1.getAbsoluteEncoder(Type.kDutyCycle);
-        this.encoder2 = new motor2.getAbsoluteEncoder(Type.kDutyCycle);
+        this.motorLeft = new motor(CLIMB_MOTOR_ONE, MotorType.kBrushless);
+        this.motorRight = new motor(CLIMB_MOTOR_TWO,MotorType.kBrushless );
+        this.encoderLeft = new motorLeft.getAbsoluteEncoder(Type.kDutyCycle);
+        this.encoderRight = new motorRight.getAbsoluteEncoder(Type.kDutyCycle);
 
+        //PID
+        this.motorLeftPIDController = motorLeft.getPIDController();
+        this.motorRightPIDController = motorRight.getPIDController();
+        motorLeftPIDController.setFeedbackDevice(encoderLeft);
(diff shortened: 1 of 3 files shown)


jamesdooley4:
@Yumnah2 - need to fix spotlessJavaCheck failures.

src/main/java/frc/team2412/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,121 @@
...
+    //private final double MOTOR_SPEED = 0.15;
+        //PID extension values
+        //PID retraction values
+        //height of chain in encoder??? units
+
+    //  INSTANCE VARIABLES
+    private final CANSparkFlex motorLeft;
+    private final CANSparkFlex motorRight;
+    private final SparkAbsoluteEncoder encoderLeft;
+    private final SparkAbsoluteEncoder encoderRight;
```
jamesdooley4, line 32: Should delete these, as the absolute encoder is only useful for measuring less than a single turn. Assuming climb rotates more than one turn, the `RelativeEncoder` returned from `motorLeft.getEncoder()` is what will need to be used.

src/main/java/frc/team2412/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,121 @@
...
+        encoderLeft = motorLeft.getAbsoluteEncoder(SparkAbsoluteEncoder.Type.kDutyCycle);
+        encoderRight = motorRight.getAbsoluteEncoder(SparkAbsoluteEncoder.Type.kDutyCycle);
+
+        //PID
+        this.motorLeftPIDController = motorLeft.getPIDController();
+        this.motorRightPIDController = motorRight.getPIDController();
+        motorLeftPIDController.setFeedbackDevice(encoderLeft);
+        motorRightPIDController.setFeedbackDevice(encoderRight);
+
+        //motor configuration
```
jamesdooley4, line 50: Need to also reset to factory default, set inverted mode (or not), set current limits, and burn configuration to flash so it persists across power loss. See https://github.com/robototes/ChargedUp2023/blob/f25191718390d11df9ddda789a0f0d5c35686734/src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java#L124 for an example

src/main/java/frc/team2412/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,121 @@
...
+    public double getSpeed() {
+        return motorLeft.get();
+    }
+
+    public void setMotorUp() {
+        setMotor(EXTEND_SPEED);
+    }
+    
+    public void setMotorDown() {
+        setMotor(-RETRACT_SPEED);
```
jamesdooley4, line 77: `RETRACT_SPEED` is already defined as negative above, so shouldn't be inverted again here.

src/main/java/frc/team2412/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,121 @@
...
+    //set motor position
+    //get+set motor speed
+    //stop arm
+    public double getSpeed() {
+        return motorLeft.get();
+    }
+
+    public void setMotorUp() {
+        setMotor(EXTEND_SPEED);
+    }
```
jamesdooley4, line 74: FYI that when climbing at the edges (where the chain will be much higher on one side of the robot than the other), the left & right climb motors may need to go up to different heights. Fine to assume climbing in the middle with both sides going to the same level, but the controls may need to be more complex later to handle cases where left & right sides need to go up to different heights.

src/main/java/frc/team2412/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,121 @@
...
+    public void setMotorDown() {
+        setMotor(-RETRACT_SPEED);
+    }
+    
+    public void setMotor(double speed) {
+        motorLeft.set(speed);
+        motorRight.set(speed);
+    }
+
+    public void stopArm() {
```
jamesdooley4, line 85: Rename for climb. Maybe just `stop()`

src/main/java/motorRight/getAbsoluteEncoder.java
```diff
@@ -0,0 +1,5 @@
+package motorRight;
+
+public class getAbsoluteEncoder {
```
jamesdooley4, line 3: Delete extraneous file

src/main/java/frc/team2412/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+
+    //changes the spring in climb
+    //please change rotation variable later
+    public void climb(double rotations) {
+        motorLeftPIDController.setReference(rotations, ControlType.kPosition);
+        motorRightPIDController.setReference(rotations, ControlType.kPosition);
+    }
```
jamesdooley4, line 113: I'm not sure what happens when using both the built-in PID controller (by setting the reference value here) and also setting a raw power (the `.set()` commands above). Having the ability to specify a raw power when retracting the climb arms has been useful in the past. But for extension, it'd likely be better to just use `.setReference()` so the units are in "amount being extended" instead of "speed of extension".

jamesdooley4 (changes requested):

src/main/java/frc/team2412/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,121 @@
...
+        return encoderLeft.getPosition();
+    }
+    
+    public double encoderRightPosition() {
+        return encoderRight.getPosition();
+    }
+
+    //calculate height climbed
+    //reset encoder to 0 if reset == true
+    public void resetEncoder(boolean reset) {
```
TAKBS2412, line 100: What's the intented purpose of the `reset` parameter here? Could omit it and only call `resetEncoder()` when the encoder needs to be reset.

src/main/java/frc/team2412/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,121 @@
...
+        if (reset) {
+            encoderLeft.setPositionConversionFactor(0);
+            encoderRight.setPositionConversionFactor(0);
```
TAKBS2412, line 103: Since this factor is multiplied by the raw value from the encoder ([docs](https://codedocs.revrobotics.com/java/com/revrobotics/absoluteencoder#setPositionConversionFactor(double))), this will make the encoder always return a position of 0, even after moving.

TAKBS2412 (changes requested):

iamawesomecat:
no more climb :C

PR #18, opened 2024-01-27
Add x-wheels, heading correction, and super basic PathPlanner support

jbko6:
I plan to add a toggle for x-wheels on the controller and a toggle for heading correction in Shuffleboard in the future.


jbko6:
grinding out those lines changed 💪 
image

.pathplanner/settings.json
TAKBS2412, line 1: Does this file and navgrid.json need to be included, or should we add to gitignore like the other json files?

TAKBS2412 (approved):

jamesdooley4 (approved):

.pathplanner/settings.json
jbko6: I think it's a good idea to include that file as I believe it affects how PathPlanner generates paths and I imagine we would want to keep that consistent.

src/main/java/frc/team2412/robot/subsystems/DrivebaseSubsystem.java
```diff
@@ -20,21 +20,22 @@
...
 public class DrivebaseSubsystem extends SubsystemBase {
 
 	// SWERVE CONSTANTS (that aren't in deploy dir)
 
-	private static final double MAX_SPEED = 0.1;
+	private static final double MAX_SPEED = 2.0;
 	private static final double JOYSTICK_DEADBAND = 0.05;
-	private static final double DRIVEBASE_RADIUS = 0;
+	private static final double DRIVEBASE_RADIUS =
+			Math.hypot(8.5, 8.5); // our wheels are 8.5 inches by 8.5 inches from the center of the bot;
```
KangarooKoala, line 31: Should this be calculated from `SwerveDrive.swerveDriveConfiguration.moduleLocationsMeters` (because we have different robots)?

PR #19, opened 2024-01-27
Fix errorprone error that causes robot code to brick

jbko6:
Before this fix robot code would not work on the robot, returning a `Could not find or load main class frc.team2412.robot.Main` error.


TAKBS2412 (approved):

TAKBS2412:
Note for posterity: This didn't actually cause the robot to brick, the issue was fixed by clearing the contents of the `.gradle` directory in the user's home directory. This fixed the `NoClassDefFound` error we were encountering earlier (and tried to fix with this PR).

PR #24, opened 2024-01-31
Add support for Crane

jbko6:
Adds new YAGSL config for Crane


src/main/deploy/craneswerve/modules/backleft.json
```diff
@@ -0,0 +1,27 @@
+{
+  "location": {
+    "front": -8.5,
+    "left": 8.5
```
jamesdooley4: Crane isn't a square, so these look like placeholders.

jamesdooley4 (dismissed):

src/main/deploy/craneswerve/modules/backleft.json
jbko6: Whoops, those should be modified

jamesdooley4 (approved):

PR #26, opened 2024-02-03
Drivebase logging and config enhancement

jbko6:
MatchDashboard is meant to be a shuffleboard tab with only the necessary logging and configuration drivers need during a match. That way our co driver won't need to click around tabs as much during the match


jbko6:
what the match tab looks like right now
image

networktables.json.bck
```diff
@@ -0,0 +1,29 @@
+[
+{
```
jamesdooley4: Should delete this (again).

jamesdooley4 (dismissed):

jamesdooley4 (approved):

PR #32, opened 2024-02-03
Add support for new practice drivebase

jbko6:
This is the drivebase that completely mirrors the competition bot in terms of dimensions and motor configuration. I've renamed bonk's configurations (which were before referred to as "practice" configurations) to "bonk" configurations


jbko6:
Once we have our comp bot finished we will have 4️⃣!!!!!! robots to play around with in this repo

jamesdooley4 (approved):

PR #35, opened 2024-02-09
LEDSubsystem

Jetblackdragon:


src/main/java/frc/team2412/robot/subsystems/LEDSubsystem.java
```diff
@@ -0,0 +1,38 @@
...
+	private final PWM blinkin;
+
+	public LEDSubsystem() {
+		blinkin = new PWM(BLINKIN_LED);
+		blinkin.setSpeed(RED_LED_COLOR); // sets to red
+	}
+
+	public void setLED(double color) {
+		blinkin.setSpeed(color);
+	}
```
jamesdooley4: Should be private, as nothing outside the class should call this

src/main/java/frc/team2412/robot/subsystems/LEDSubsystem.java
```diff
@@ -0,0 +1,38 @@
...
+	private static final double RED_LED_COLOR = 0.61;
+    private static final double GREEN_LED_COLOR = 0.77;
+    private static final double BLUE_LED_COLOR = 0.87;
+    private static final double YELLOW_LED_COLOR = 0.69;
+
+	private final PWM blinkin;
+
+	public LEDSubsystem() {
+		blinkin = new PWM(BLINKIN_LED);
+		blinkin.setSpeed(RED_LED_COLOR); // sets to red
```
jamesdooley4: Would be better to use `setPulseTimeMicroseconds​` instead of `setSpeed`, to not depend on the PWM output range (see https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj/PWM.html )

src/main/java/frc/team2412/robot/subsystems/LEDSubsystem.java
```diff
@@ -0,0 +1,38 @@
...
+
+import static frc.team2412.robot.Hardware.*;
+import edu.wpi.first.wpilibj.PWM;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+
+public class LEDSubsystem extends SubsystemBase {
+	private static final double RED_LED_COLOR = 0.61;
+    private static final double GREEN_LED_COLOR = 0.77;
+    private static final double BLUE_LED_COLOR = 0.87;
+    private static final double YELLOW_LED_COLOR = 0.69;
```
jamesdooley4: See the bottom of https://www.revrobotics.com/content/docs/REV-11-1105-UM.pdf for th pulse width values to use instead of the speed values here (which only work if the PWM port was configured to be used with a SPARK MAX somehow).

src/main/java/frc/team2412/robot/subsystems/LEDSubsystem.java
jamesdooley4: Or update this to take an enum as the parameter that the class internally converts to a pulse width value for the color or pattern to use.

src/main/java/frc/team2412/robot/commands/drivebase/DriveCommand.java
jamesdooley4, line 1: Please delete this empty file

src/main/java/frc/team2412/robot/subsystems/LimelightSubsystem.java
```diff
@@ -0,0 +1,214 @@
...
+	private static final PIDController ROTATION_PID = new PIDController(8.0, 0, 0);
+
+	// meters?
+	public static final double CAMERA_MOUNT_HEIGHT = 0.1143;
+	public static final double CAMERA_ANGLE_OFFSET = 0;
+	public static final double TARGET_HEIGHT = 0.33;
+
+	public static final double GOAL_DISTANCE_FROM_TARGET = 0.7;
+	public static final double GOAL_DISTANCE_FROM_CONE = 0.3;
+	public static final double GOAL_DISTANCE_FROM_CUBE = 0.3;
```
jamesdooley4: Should delete since these aren't used in this year's game.

src/main/java/frc/team2412/robot/subsystems/LimelightSubsystem.java
```diff
@@ -0,0 +1,214 @@
...
+	String currentPoseString;
+	String targetPoseString;
+
+	// network tables
+
+	// CONSTRUCTOR !
+	public LimelightSubsystem() {
+
+		// broadcast 10.24.12.227
+
```
jamesdooley4: That's a dynamic IP address that will change once the robot is connected to the FMS. This needs to not conflict with [AprilTag PR](https://github.com/robototes/Crescendo2024/pull/41) and be between 10.TE.AM.5 (or .6, if the driver station is statically set to .5) to 10.TE.AM.20 (inclusive) per https://docs.wpilib.org/en/stable/docs/networking/networking-introduction/ip-configurations.html#on-the-field-static-configuration
Anything above .20 might be allocated elsewhere by the FMS. Typically the LL is .10

src/main/java/frc/team2412/robot/subsystems/LimelightSubsystem.java
```diff
@@ -0,0 +1,214 @@
...
+				.withSize(4, 1);
+
+		limelightTab
+				.addString("Target Pose ", this::getTargetPoseString)
+				.withPosition(0, 2)
+				.withSize(4, 1);
+	}
+
+	public void setPipeline() {
+		networkTable.getEntry("pipeline").setNumber(2);
```
jamesdooley4: Should be private, as nothing outside the class should call it

src/main/java/frc/team2412/robot/subsystems/LimelightSubsystem.java
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	public double getDistanceFromTarget() {
+
+		// 1.3 at 1.9
+		// 1.0 at 1.3
+
+		double angleToTarget = getVerticalOffset();
+		// return (TARGET_HEIGHT - CAMERA_MOUNT_HEIGHT) / Math.tan(CAMERA_ANGLE_OFFSET + angleToTarget);
+
```
jamesdooley4: If this is important to keep, please add a comment explaining why. If not, please delete.

src/main/java/frc/team2412/robot/subsystems/LimelightSubsystem.java
```diff
@@ -0,0 +1,214 @@
...
+		return (TARGET_HEIGHT - CAMERA_MOUNT_HEIGHT)
+				/ (CAMERA_ANGLE_OFFSET + Math.tan(Units.degreesToRadians(getVerticalOffset())));
+	}
+
+	public double getDistanceFromTargetTheSecond() {
+
```
jamesdooley4: Consider returning `Measure<Distance>` and using `Inches.of(...)` below so the caller can use any units it wants.

src/main/java/frc/team2412/robot/subsystems/LimelightSubsystem.java
```diff
@@ -0,0 +1,214 @@
...
+		return (getDistanceFromTarget() <= GOAL_DISTANCE_FROM_TARGET);
+	}
+
+	public Command getWithinDistance(DrivebaseSubsystem drivebaseSubsystem) {
+		final DoubleSupplier returnTurn = () -> turnPowerExp();
+		final DoubleSupplier returnZero = () -> 0.0;
+		Command moveCommand;
+		Pose2d targetPose;
+		if (hasTargets()) {
+			System.out.println("has targets");
```
jamesdooley4: Please delete (or comment) the println before completing the PR

jamesdooley4 (changes requested):

src/main/java/frc/team2412/robot/commands/drivebase/DriveCommand.java
jamesdooley4: The empty DriveCommand.java is still there...

Jetblackdragon: fixed

PR #36, opened 2024-02-10
Add FF to SparkPIDWidget and fix bug with name

### LINKED ISSUE #34: Add feedforward property to SparkPIDWidget
Should use `setFF` and `getFF` here to enable easy tweaking of the feedforward gain constant:
https://github.com/robototes/Crescendo2024/blame/e5d2fd6a6d25049a37705ddc76b2a92b5178fcd9/src/main/java/frc/team2412/robot/util/SparkPIDWidget.java#L24

jbko6:
resolves #34

FF value getter/setter in Shuffleboard is not loaded automatically into the layout. It can be added by opening the NetworkTables table on the left and dragging the value over as an entry. This is a limitation with Shuffleboard itself and would need to be fixed with a plugin which would be pretty janky.

### DIFF
--- a/src/main/java/frc/team2412/robot/util/SparkPIDWidget.java
+++ b/src/main/java/frc/team2412/robot/util/SparkPIDWidget.java
@@ -9,10 +9,10 @@ public class SparkPIDWidget implements NTSendable {
 
 	public final SparkPIDController controller;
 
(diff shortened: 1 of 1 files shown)


iamawesomecat:
I also have the bug fixed on my PR, want me to get rid of it?

jamesdooley4 (approved):

PR #37, opened 2024-02-10
Add new alignment command to Drivebase

jbko6:
This is anticipating the need for apriltags/launcher systems needing to rotate the drivebase to align to a target. The way this is written should also allow for launching while driving.


jbko6:
This PR also includes the Crane RoboRIO / MAC address swap that needs to be added at some point

jbko6:
also makes x wheels setting persistent across code deployments which should help when we're testing and have to keep turning it off over and over

jamesdooley4 (approved):

bobtseattle (approved):
LGTM

Issue #29, opened 2024-02-03
LauncherSubsystem null error


jbko6:
When LauncherSubsystem is enabled and run in simulation a null error results

`Error at frc.team2412.robot.subsystems.LauncherSubsystem.getAngle(LauncherSubsystem.java:119): Unhandled exception: java.lang.NullPointerException: 
Cannot invoke "com.revrobotics.SparkAbsoluteEncoder.getPosition()" because "this.launcherAngleEncoder" is null`

The issue is in line 50 of LauncherSubsystem `.add("Launcher angle", getAngle())`
The encoder is not yet initialized when `getAngle()` is called

iamawesomecat:
fixed :D

PR #38, opened 2024-02-10
Intake Sensors + Controls

Yumnah2:


src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
```diff
@@ -33,6 +34,11 @@ public class IntakeSubsystem extends SubsystemBase {
 
 	private final CANSparkFlex feederMotor;
 
+	// Sensors
+	private final DigitalInput intakeSensor;
```
jamesdooley4: The robot currently doesn't have an intake sensor, so should remove this & the getter method below

jamesdooley4 (changes requested):

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
jamesdooley4: Rest of the code looks fine.  Nice work!

eddiert1239: Remove only the intake sensor and it's getter method, or the rest of them as well?

src/main/java/frc/team2412/robot/Hardware.java
```diff
@@ -20,7 +20,13 @@ public class Hardware {
 	public static final int INTAKE_MOTOR_LEFT = 42;
 	public static final int INTAKE_MOTOR_RIGHT = 43;
 
-	public static final int INDEX_MOTOR = 44;
+	public static final int INDEX_MOTOR_LOWER = 44;
+	public static final int INDEX_MOTOR_UPPER = 44;
```
TAKBS2412: These should be two different IDs.

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
jamesdooley4: Just the intake sensor. The other sensors are good & should stay

TAKBS2412 (changes requested):

eddiert1239:
Controls right now should be:
x button intakes in until the feeder sensor returns true, then all motors stop
b button stops all motors (intake index feeder)
y button intakes out while the button is held

src/main/java/frc/team2412/robot/Hardware.java
eddiert1239: That wasn't very smart of us!

src/main/java/frc/team2412/robot/commands/intake/FeederOutCommand.java
```diff
@@ -15,4 +15,9 @@ public FeederOutCommand(IntakeSubsystem intakeSubsystem) {
 	public void initialize() {
 		intakeSubsystem.feederOut();
 	}
+
+	@Override
+	public boolean isFinished() {
+		return false;
+	}
```
jamesdooley4: The feeder out, intake out & intake stop commands should all continue to return 'true' from isFinished (the defafult behavior, so can delete the override `isFinished` methods)

jamesdooley4 (changes requested):

src/main/java/frc/team2412/robot/commands/intake/FeederInCommand.java
```diff
@@ -15,14 +15,4 @@ public FeederInCommand(IntakeSubsystem intakeSubsystem) {
...
-
-	@Override
-	public void end(boolean interrupted) {
-		intakeSubsystem.feederStop();
-	}
-
-	@Override
-	public boolean isFinished() {
-		return intakeSubsystem.feederSensor();
-	}
```
TAKBS2412: This part should be kept (same for `IndexInCommand` and `IntakeInCommand`) - what James and I meant was deleting `isFinished()` from the commands where it just returns `false`, like `FeederOutCommand`. Sorry for any confusion earlier.

TAKBS2412 (changes requested):

src/main/java/frc/team2412/robot/commands/intake/FeederInCommand.java
jamesdooley4: Looking at [Command::isFinished](https://github.com/wpilibsuite/allwpilib/blob/da3abade83ada38d387d7ab0633f0914a44ff9d1/wpilibNewCommands/src/main/java/edu/wpi/first/wpilibj2/command/Command.java#L62), it returns `false` by default. I thought it returned `true`. So, the `isFinished()` override needs to be kept, and just updated to return `true` so the diagnostic command works correctly. Sorry about the confusion.

jamesdooley4 (dismissed):

src/main/java/frc/team2412/robot/commands/intake/FeederInCommand.java
TAKBS2412: I'm fine with making commands like `IntakeStopCommand` and `IntakeOutCommand` return `true`, but for the "in" commands (`IndexInCommand`, `IntakeInCommand`, and `FeederInCommand`) I think we should keep `isFinished()` as is (checking the sensors) since the intent was to make the command run until a note is detected, which is additional logic I don't think we should compromise for a diagnostic command.

To keep the diagnostic command from being stuck waiting for a note, I'd suggest adding [`withTimeout()`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/Command.html#withTimeout(double)) to make the in commands only run for a given period of time. (I could also see some value in keeping in commands that use the sensors, just to verify that the sensor hardware is working correctly).

Sorry for all the back and forth :(

TAKBS2412 (changes requested):

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
```diff
@@ -125,4 +137,13 @@ public void feederOut() {
 	public void feederStop() {
 		feederMotor.set(0);
 	}
+
+	// sensor methods
+	public boolean indexSensor() {
```
kirbt: It would be nice to log these sensor getter methods for testing/debugging purposes

kirbt: also the sensor method names are a bit iffy

src/main/java/frc/team2412/robot/commands/intake/FeederInCommand.java
jamesdooley4: Seems reasonable.

TAKBS2412 (approved):

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
```diff
@@ -53,16 +59,32 @@ public class IntakeSubsystem extends SubsystemBase {
 					.withSize(1, 1)
 					.getEntry();
 
+	private final GenericEntry getIndexSensorEntry =
+			Shuffleboard.getTab("Intake")
+					.add("Index Sensor - ", getIndexSensor())
```
TAKBS2412: The `getIndexSensor()` call causes crashes (also in the `getFeederSensorEntry` below) because the index sensor hasn't been initialized yet. See #29 for a similar error that launcher subsystem ran into, and how to fix it (note you will have to update the entry's value in `periodic()`).

Alternatively, you can initialize the entries after the `DigitalInput`, and use something like [this](https://github.com/robototes/ChargedUp2023/blob/main/src/main/java/frc/team2412/robot/subsystems/VisionSubsystem.java#L165) to repeatedly call `getIndexSensor()` without using `periodic()`.

TAKBS2412 (changes requested):

src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -61,6 +83,28 @@ private void bindDrivebaseControls() {
...
+	// intake controls
+	private void bindIntakeControls() {
+		// later on set default command to check sensing notes
+		CommandScheduler.getInstance()
+				.setDefaultCommand(s.intakeSubsystem, new IntakeStopCommand(s.intakeSubsystem));
+		codriveIntakeInButton.onTrue(
+				Commands.race(
+						new IntakeInCommand(s.intakeSubsystem),
+						new IndexInCommand(s.intakeSubsystem),
+						new FeederInCommand(s.intakeSubsystem)));
```
TAKBS2412: Note that these and the parallel command groups below will throw an exception:

```
Error at frc.team2412.robot.Controls.bindIntakeControls(Controls.java:92): Unhandled exception: java.lang.IllegalArgumentException: Multiple commands in a parallel composition cannot require the same subsystems
        at edu.wpi.first.wpilibj2.command.ParallelRaceGroup.addCommands(ParallelRaceGroup.java:53)
        at edu.wpi.first.wpilibj2.command.ParallelRaceGroup.<init>(ParallelRaceGroup.java:35)
        at edu.wpi.first.wpilibj2.command.Commands.race(Commands.java:228)
        at frc.team2412.robot.Controls.bindIntakeControls(Controls.java:92)
        at frc.team2412.robot.Controls.<init>(Controls.java:69)
        at frc.team2412.robot.Robot.robotInit(Robot.java:67)
        at edu.wpi.first.wpilibj.TimedRobot.startCompetition(TimedRobot.java:107)
        at edu.wpi.first.wpilibj.RobotBase.runRobot(RobotBase.java:366)
        at edu.wpi.first.wpilibj.RobotBase.lambda$startRobot$0(RobotBase.java:438)
        at java.base/java.lang.Thread.run(Thread.java:833)
```

src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -61,6 +83,28 @@ private void bindDrivebaseControls() {
 		driveController.rightStick().onTrue(new InstantCommand(s.drivebaseSubsystem::toggleXWheels));
 	}
 
+	// intake controls
+	private void bindIntakeControls() {
+		// later on set default command to check sensing notes
+		CommandScheduler.getInstance()
+				.setDefaultCommand(s.intakeSubsystem, new IntakeStopCommand(s.intakeSubsystem));
```
TAKBS2412: Because `IntakeOutCommand` ends immediately, the intake motors will only run briefly before `IntakeStopCommand` makes them all stop. I think not having a default command is ok for now especially since this one is just a placeholder.

TAKBS2412 (changes requested):

eddiert1239:
### Controls now:
Drive controller:
x - in
y - out
b - stop
Codrive controller:
dpad left - in
dpad up - out
dpad right - stop

src/main/java/frc/team2412/robot/commands/intake/AllOutCommand.java
```diff
@@ -0,0 +1,25 @@
+package frc.team2412.robot.commands.intake;
+
+import edu.wpi.first.wpilibj2.command.Command;
+import frc.team2412.robot.subsystems.IntakeSubsystem;
+
+public class AllOutCommand extends Command {
```
bobtseattle: Out seems ambiguous.  It might spit the note out the bottom or out the launcher.

bobtseattle (changes requested):

TAKBS2412 (dismissed):

bobtseattle (approved):
LGTM

PR #40, opened 2024-02-13
Add FullTargetCommand and CSV intepreter for launcher data

jbko6:


.gitignore
TAKBS2412, line 1: I think this file can be reverted?

src/main/deploy/craneswerve/controllerproperties.json
TAKBS2412, line 1: ? This file is listed as a binary file, but has the json file extension, which is not a binary file format.

jamesdooley4: From https://github.com/robototes/Crescendo2024/blob/a751e140b7bd87c8a20eb60a92bb7d792a7a2d69/src/main/deploy/craneswerve/controllerproperties.json it looks like it's now 98 unknown (null?) characters, so got corrupted somehow

src/main/deploy/launcher_data.csv
```diff
@@ -0,0 +1,8 @@
+# test comment
+
+# DISTANCE (meters), ANGLE (degrees), RPM (velocity)
+
+0, 0, 0
+1, 10, 2000
+2, 20, 2200
+5, 40, 3000
```
jamesdooley4: Better to start with more realistic data, where the angle is lower at a larger distance, and distance of '0' has valid values (assuming that's launching from the subwoofer).

src/main/java/frc/team2412/robot/commands/diagnostic/LauncherDiagnosticCommand.java
```diff
@@ -15,13 +14,8 @@ public LauncherDiagnosticCommand(LauncherSubsystem launcherSubsystem) {
...
-				new WaitCommand(2),
-				new SetAngleCommand(launcherSubsystem, 90),
-				new WaitCommand(1),
-				new SetAngleCommand(launcherSubsystem, Angle),
-				new SetLaunchSpeedCommand(launcherSubsystem, 100),
-				new WaitCommand(5),
-				new StopLauncherCommand(launcherSubsystem));
+				new SetAngleCommand(launcherSubsystem, () -> 45).withTimeout(2),
+				new SetAngleCommand(launcherSubsystem, () -> Angle).withTimeout(2),
+				new SetLaunchSpeedCommand(launcherSubsystem, () -> 100).withTimeout(2));
```
jamesdooley4: Having a small wait before the SetAngleCommand calls is helpful to make it easier to check the robot physically go to the expected angle.

jamesdooley4: And having a second angle than 45 degrees ensures this moves the pivot if the arm starts at 45 degrees

src/main/java/frc/team2412/robot/commands/launcher/SetAngleLaunchCommand.java
```diff
@@ -2,14 +2,16 @@
 
 import edu.wpi.first.wpilibj2.command.SequentialCommandGroup;
 import frc.team2412.robot.subsystems.LauncherSubsystem;
+import java.util.function.DoubleSupplier;
 // this command can be used as a preset in controls, allowing the user to input a speed and angle
 // value when they keybind it multiple times.
 
 public class SetAngleLaunchCommand extends SequentialCommandGroup {
-	private double launcherSpeed;
```
jamesdooley4: Re-reading this, it would be better as a ParallelCommandGroup (instead of Sequential) so the launcher can speedup while the angle motor is getting the arm into position.

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
```diff
@@ -0,0 +1,73 @@
...
+		launcherData =
+				LauncherDataLoader.fromCSV(
+						FileSystems.getDefault()
+								.getPath(Filesystem.getDeployDirectory().getPath(), "launcher_data.csv"));
+	}
+
+	@Override
+	public void initialize() {
+		CommandScheduler.getInstance().schedule(setAngleCommand);
+		CommandScheduler.getInstance().schedule(setLaunchSpeedCommand);
```
jamesdooley4: Using `SetAngleLaunchCommand` would let the angle & launch speed be both change at the same time, saving time

jamesdooley4: Or don't schedule the commands at all & work directly on the subsystems (which would behave better, per below comment).

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
```diff
@@ -0,0 +1,73 @@
...
+	public void execute() {
+		robotPose = drivebaseSubsystem.getPose();
+		relativeSpeaker = robotPose.relativeTo(SPEAKER_POSE);
+		yawAngle =
+				Rotation2d.fromRadians(
+						Math.atan2(relativeSpeaker.getY(), relativeSpeaker.getX()) + Math.PI);
+		distance = relativeSpeaker.getTranslation().getNorm();
+		dataPoint = launcherData.get(distance);
+		rpm = dataPoint.rpm;
+		angle = dataPoint.angle;
```
jamesdooley4: This likely won't do anything, as the SetAngleCommand ends as soon as a previously set angle value is reached. It'd be better for this command to just update the launcher speed & angle targets in its own `execute()` function, and not re-use other commands.

src/main/java/frc/team2412/robot/commands/launcher/SetAngleCommand.java
```diff
@@ -3,26 +3,29 @@
...
 		this.launcherSubsystem = launcherSubsystem;
 		this.launcherAngle = angle;
 		addRequirements(launcherSubsystem);
 	}
 
 	@Override
-	public void initialize() {
-		launcherSubsystem.setAngle(launcherAngle);
+	public void execute() {
+		launcherSubsystem.setAngle(launcherAngle.getAsDouble());
```
jamesdooley4: Consider keeping this as `initialize()` to avoid callers having to check if the command has ended before they update the value in the DoubleSupplier (to avoid bugs where they don't handle the command ending since it reached the target angle).

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
```diff
@@ -0,0 +1,73 @@
...
+			LauncherSubsystem launcherSubsystem, DrivebaseSubsystem drivebaseSubsystem) {
+		this.launcherSubsystem = launcherSubsystem;
+		this.drivebaseSubsystem = drivebaseSubsystem;
+		setLaunchSpeedCommand = new SetLaunchSpeedCommand(launcherSubsystem, () -> rpm);
+		setAngleCommand = new SetAngleCommand(launcherSubsystem, () -> angle);
+		rotateToAngle = drivebaseSubsystem.rotateToAngle(() -> yawAngle, false);
+		launcherData =
+				LauncherDataLoader.fromCSV(
+						FileSystems.getDefault()
+								.getPath(Filesystem.getDeployDirectory().getPath(), "launcher_data.csv"));
```
jamesdooley4: This is likely slow & noisy (in terms of impact on the Java garbage collector due to the string operations in the parser), so best to do infrequently (ideally just once). I'd recommend making `launcherData` static, so this is only run once when the first instance of this class is created & future instances just use the same data.

jamesdooley4 (changes requested):

src/main/java/frc/team2412/robot/util/LauncherDataLoader.java
```diff
@@ -0,0 +1,117 @@
...
+			}
+
+			debugWriter.newLine();
+			debugWriter.append("Done serializing CSV");
+			debugWriter.newLine();
+			debugWriter.close();
+			System.out.println("Done serializing CSV");
+
+			return map;
+		} catch (IOException e) {
```
bobtseattle, line 113: Is it worth considering instead of returning null, we default to a single, hardcoded line (maybe from the subwoofer) in the off chance the file gets corrupted so it can shoot from at least one spot (or do we need two to interpolate)?

bobtseattle (commented):
One thought about making sure we have a fallback if things go wrong with a bad or corrupted file.

src/main/deploy/craneswerve/controllerproperties.json
KangarooKoala: I ran `fhx` locally and it is null characters:
```
           00 01 02 03 04 05 06 07 08 09 0A 0B 0C 0D 0E 0F

00000000   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
00000010   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
00000020   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
00000030   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
00000040   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
00000050   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
00000060   00 00                                            ..
```
`git blame` says the change was made in 30fbcbd8, by the way.

src/main/java/frc/team2412/robot/commands/diagnostic/LauncherDiagnosticCommand.java
```diff
@@ -1,10 +1,9 @@
 package frc.team2412.robot.commands.diagnostic;
 
+import edu.wpi.first.units.Angle;
```
KangarooKoala: What's the import for? (And should the variable be changed to lowercase to prevent intellisense from doing something like this again?)

src/main/java/frc/team2412/robot/util/LauncherDataLoader.java
```diff
@@ -0,0 +1,116 @@
...
+				new InterpolatingTreeMap<>(InverseInterpolator.forDouble(), LauncherDataPoint.INTERPOLATOR);
+
+		try {
+			BufferedReader reader = Files.newBufferedReader(path);
+			BufferedWriter debugWriter =
+					Files.newBufferedWriter(
+							FileSystems.getDefault()
+									.getPath(Filesystem.getOperatingDirectory().getPath(), "logs/CSVInterpreter.log"),
+							StandardOpenOption.CREATE,
+							StandardOpenOption.WRITE);
```
KangarooKoala: I'd recommend using a try-with-resources (`try (BufferedReader reader = ...; BufferedWriter debugWriter = ...) { ... } catch { ... }`) to ensure these get closed properly, even in the case of an exception. (And also remove the manual `close()` call towards the end- It's documented that repeated `close()` calls won't do anything bad, but it's still better to avoid it)

src/main/java/frc/team2412/robot/util/LauncherDataLoader.java
KangarooKoala: The [InterpolatingTreeMap](https://github.com/wpilibsuite/allwpilib/blob/main/wpimath/src/main/java/edu/wpi/first/math/interpolation/InterpolatingTreeMap.java) handles the past-the-edge cases, including if there's just one key. (If there's no keys, it returns null)

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
```diff
@@ -0,0 +1,92 @@
...
+		intakeSubsystem.feederStop();
+	}
+
+	@Override
+	public void execute() {
+		Pose2d robotPose = drivebaseSubsystem.getPose();
+		Pose2d relativeSpeaker = robotPose.relativeTo(SPEAKER_POSE);
+		yawTarget =
+				Rotation2d.fromRadians(
+						Math.atan2(relativeSpeaker.getY(), relativeSpeaker.getX()) + Math.PI);
```
KangarooKoala, line 69: Would it be clearer to find the speaker pose relative to the robot pose (instead of finding robot pose relative to speaker pose and adding pi)?

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
```diff
@@ -0,0 +1,92 @@
...
+	public FullTargetCommand(
+			LauncherSubsystem launcherSubsystem,
+			IntakeSubsystem intakeSubsystem,
+			DrivebaseSubsystem drivebaseSubsystem,
+			BooleanSupplier launch) {
+		this.launcherSubsystem = launcherSubsystem;
+		this.intakeSubsystem = intakeSubsystem;
+		this.drivebaseSubsystem = drivebaseSubsystem;
+		this.launch = launch;
+		yawAlignmentCommand = drivebaseSubsystem.rotateToAngle(() -> yawTarget, false);
```
KangarooKoala, line 52: Is there a reason to use an intermediate variable that's updated in `execute()` instead of having the lambda calculate the angle? (You can use `() -> { ...; return value; }` if you need multiple statements within the lambda body)

jbko6:
this crashes if you enable launcher currently, working on a fix

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
```diff
@@ -1,73 +1,93 @@
...
-		setLaunchSpeedCommand = new SetLaunchSpeedCommand(launcherSubsystem, () -> rpm);
-		setAngleCommand = new SetAngleCommand(launcherSubsystem, () -> angle);
-		rotateToAngle = drivebaseSubsystem.rotateToAngle(() -> yawAngle, false);
-		launcherData =
-				LauncherDataLoader.fromCSV(
-						FileSystems.getDefault()
-								.getPath(Filesystem.getDeployDirectory().getPath(), "launcher_data.csv"));
+		this.launch = launch;
+		yawAlignmentCommand = drivebaseSubsystem.rotateToAngle(() -> yawTarget, false);
 	}
```
jamesdooley4: This needs to require the intake & launcher subsystem so it doesn't fight with another simultaneous command that's also changing things in those subsystems.

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.function.BooleanSupplier;
+
+public class FullTargetCommand extends Command {
+
+	private static final InterpolatingTreeMap<Double, LauncherDataPoint> LAUNCHER_DATA =
+			LauncherDataLoader.fromCSV(
+					FileSystems.getDefault()
+							.getPath(Filesystem.getDeployDirectory().getPath(), "launcher_data.csv"));
+	private static final float YAW_TARGET_VIBRATION_TOLERANCE = 3; // degrees
+	private final Pose2d SPEAKER_POSE = new Pose2d(0.0, 5.55, Rotation2d.fromRotations(0));
```
KangarooKoala: When will handling of different alliances be added?

jamesdooley4 (changes requested):

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
```diff
@@ -0,0 +1,93 @@
...
+		if (launch.getAsBoolean()) {
+			intakeSubsystem.feederIn();
+		} else {
+			intakeSubsystem.feederStop();
+		}
+
+		if (MathUtil.isNear(
+				yawTarget.getDegrees(),
+				drivebaseSubsystem.getPose().getRotation().getDegrees(),
+				YAW_TARGET_VIBRATION_TOLERANCE)) {
```
KangarooKoala:
```suggestion
		if (MathUtil.isNear(
				yawTarget.getDegrees(),
				drivebaseSubsystem.getPose().getRotation().getDegrees(),
				YAW_TARGET_VIBRATION_TOLERANCE,
				0,
				360)) {
```
This handles the angle wraparound (using the [`MathUtil.isNear(expected, actual, tolerance, min, max)`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/math/MathUtil.html#isNear(double,double,double,double,double)) overload.

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
```diff
@@ -0,0 +1,93 @@
...
+import java.nio.file.FileSystems;
+import java.util.function.BooleanSupplier;
+
+public class FullTargetCommand extends Command {
+
+	private static final InterpolatingTreeMap<Double, LauncherDataPoint> LAUNCHER_DATA =
+			LauncherDataLoader.fromCSV(
+					FileSystems.getDefault()
+							.getPath(Filesystem.getDeployDirectory().getPath(), "launcher_data.csv"));
+	private static final float YAW_TARGET_VIBRATION_TOLERANCE = 3; // degrees
```
KangarooKoala: Minor: Why is this a float and not a double?

src/main/java/frc/team2412/robot/commands/launcher/SetLaunchSpeedCommand.java
```diff
@@ -11,7 +11,6 @@ public class SetLaunchSpeedCommand extends Command {
 	public SetLaunchSpeedCommand(LauncherSubsystem launcherSubsystem, double speed) {
 		launcherSpeed = speed;
 		this.launcherSubsystem = launcherSubsystem;
-		addRequirements(launcherSubsystem);
```
KangarooKoala: Why was this requirement removed?

src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -67,6 +69,17 @@ public Controls(Subsystems s) {
...
+		if (DRIVEBASE_ENABLED && LAUNCHER_ENABLED && INTAKE_ENABLED) {
+			// temporary controls, not sure what drive team wants
+			driveController
+					.leftBumper()
+					.whileTrue(
+							new FullTargetCommand(
+									s.launcherSubsystem,
+									s.intakeSubsystem,
+									s.drivebaseSubsystem,
+									() -> driveController.rightBumper().getAsBoolean()));
```
KangarooKoala: As a side note, `driveController.rightBumper()` is a `Trigger`, which implements `DoubleSupplier`, so you can pass `driveController.rightBumper()` instead of `() -> driveController.rightBumper().getAsBoolean()`.

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
```diff
@@ -0,0 +1,97 @@
...
+		if (launch.getAsBoolean()) {
+			intakeSubsystem.feederIn();
+		} else {
+			intakeSubsystem.feederStop();
+		}
+
+		if (MathUtil.isNear(
+						yawTarget.getDegrees(),
+						drivebaseSubsystem.getPose().getRotation().getDegrees(),
+						YAW_TARGET_VIBRATION_TOLERANCE)
```
KangarooKoala:
```suggestion
		if (MathUtil.isNear(
						yawTarget.getDegrees(),
						drivebaseSubsystem.getPose().getRotation().getDegrees(),
						YAW_TARGET_VIBRATION_TOLERANCE,
						0,
						360)
```
Without this change, it won't handle angle wraparound.

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
```diff
@@ -0,0 +1,97 @@
...
+
+		addRequirements(launcherSubsystem, intakeSubsystem);
+	}
+
+	@Override
+	public void initialize() {
+		CommandScheduler.getInstance().schedule(yawAlignmentCommand);
+		intakeSubsystem.feederStop();
+		Robot robot = Robot.getInstance();
+		controls = robot.controls;
```
KangarooKoala: Would it be cleaner to pass the `Controls` instance in the constructor (or even just pass a `DoubleSupplier` with the rumble setter)?

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
jbko6: it's a bad habit

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
jbko6: I dont think we need to add handling for this as the field is symmetrical along the center line this year

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
jbko6: Personally I think it's more readable this way as all the updates to robot state are in the same part of the code (execute method)

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
jbko6: When I tried doing Tuesday that it didn't work ¯\_(ツ)_/¯
Might be worth investigating again later but for now I know this works with the real robot

src/main/java/frc/team2412/robot/commands/launcher/SetLaunchSpeedCommand.java
jbko6: It was to address James' comment that parallel command groups cannot have two requirements for the same subsystem. Since SetLaunchSpeedCommand ends immediately I thought it was in less need of requiring LauncherSubsystem

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
KangarooKoala: What's the plan for the coordinate system when on blue alliance vs when on red alliance?

jamesdooley4: Specifically, the Apriltags will have different positions / poses / etc.

jbko6: nvm i was confused by how PathPlanner was handling alliance switching

src/main/java/frc/team2412/robot/commands/launcher/SetLaunchSpeedCommand.java
jamesdooley4: Composing multiple commands in parallel that use the same subsystem isn't going to work. Rather than removing the requirement (which means sequential use elsewhere is likely going to end up with two commands telling a subsystem to do different things), it'd be better to add a new command that does the combination that's needed. To address the proliferation of tiny command classes, consider adding command helpers to the subsystem (like `DrivebaseSubsystem::rotateToAngle` did) rather than small files like this.
From https://github.com/wpilibsuite/allwpilib/blob/6afff996401343c3bfcc46a40a40abda429e5077/wpilibNewCommands/src/main/java/edu/wpi/first/wpilibj2/command/ParallelCommandGroup.java#L51C1-L54C8
```
      if (!Collections.disjoint(command.getRequirements(), m_requirements)) {
        throw new IllegalArgumentException(
            "Multiple commands in a parallel composition cannot require the same subsystems");
      }
```

jbko6: @iamawesomecat Can you address this in your PR? I think this fix would be getting away from the original purpose of this PR. I am going to remove my current fix from this PR.

iamawesomecat: This was fixed in my PR as far as im aware.

jamesdooley4 (approved):

PR #41, opened 2024-02-14
Add AprilTags robot localization

KangarooKoala:
Some other miscellaneous changes:
* Make stuff work without drivebase enabled (so that I could test robot localization without drivebase, though I'm still not sure about the exact cases we want to handle)
* Prevent spotless from processing networktables.json and networktables.json.bck
* Let the generated Javadocs reference the wpilib docs (I looked at them once to check an `@see` specification)
* Add SendablePose3d to help with the dashboard display (though I haven't actually ran Shuffleboard with the robot code yet)

### DIFF
--- a/src/main/java/frc/team2412/robot/Hardware.java
+++ b/src/main/java/frc/team2412/robot/Hardware.java
@@ -1,7 +1,11 @@
 package frc.team2412.robot;
 
 public class Hardware {
-	// this is where we will ID the hardware, like motors n stuff
+	// apriltags stuff
+	public static final String PHOTON_CAM = "Arducam_OV9281_USB_Camera";
+	public static final String PHOTON_IP = "10.24.12.51";
+
+	// motor IDs:
 
 	// drive devices [1 - 19]
 
--- a/src/main/java/frc/team2412/robot/Hardware.java
+++ b/src/main/java/frc/team2412/robot/Hardware.java
@@ -3,7 +3,7 @@ package frc.team2412.robot;
 public class Hardware {
 	// apriltags stuff
 	public static final String PHOTON_CAM = "Arducam_OV9281_USB_Camera";
-	public static final String PHOTON_IP = "10.24.12.51";
+	public static final String PHOTON_IP = "10.24.12.12";
 
 	// motor IDs:
 
--- a/src/main/java/frc/team2412/robot/sensors/AprilTagsProcessor.java
+++ b/src/main/java/frc/team2412/robot/sensors/AprilTagsProcessor.java
@@ -1,0 +1,150 @@
+package frc.team2412.robot.sensors;
+
+import edu.wpi.first.apriltag.AprilTagFieldLayout;
+import edu.wpi.first.apriltag.AprilTagFields;
+import edu.wpi.first.math.VecBuilder;
+import edu.wpi.first.math.Vector;
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.math.geometry.Pose3d;
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.math.geometry.Rotation3d;
+import edu.wpi.first.math.geometry.Transform3d;
+import edu.wpi.first.math.numbers.N3;
+import edu.wpi.first.math.util.Units;
+import edu.wpi.first.networktables.NetworkTableEvent;
+import edu.wpi.first.networktables.NetworkTableInstance;
+import edu.wpi.first.wpilibj.shuffleboard.Shuffleboard;
+import edu.wpi.first.wpilibj.shuffleboard.ShuffleboardTab;
+import edu.wpi.first.wpilibj.smartdashboard.FieldObject2d;
+import frc.team2412.robot.Hardware;
+import frc.team2412.robot.Robot;
+import frc.team2412.robot.util.DrivebaseWrapper;
+import frc.team2412.robot.util.SendablePose3d;
+import java.util.EnumSet;
+import java.util.Optional;
+import org.photonvision.EstimatedRobotPose;
+import org.photonvision.PhotonCamera;
+import org.photonvision.PhotonPoseEstimator;
+import org.photonvision.PhotonPoseEstimator.PoseStrategy;
+import org.photonvision.targeting.PhotonPipelineResult;
+
+/**
+ * All 3D poses and transforms use the NWU (North-West-Up) coordinate system, where +X is
+ * north/forward, +Y is west/left, and +Z is up. On the field, this is based on the blue driver
+ * station. (+X is forward from blue driver station, +Y is left, and +Z is up. Additionally, the
+ * origin is the right corner of the blue driver station- All coordinates should be non-negative.)
+ *
+ * <p>Rotations follow the right-hand rule- If the thumb of a right hand is pointed in a positive
+ * axis, the fingers curl in the direction of positive rotation. Alternatively, the rotation is CCW+
+ * looking into the positive axis.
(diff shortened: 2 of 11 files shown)


src/main/java/frc/team2412/robot/Hardware.java
```diff
@@ -1,7 +1,11 @@
 package frc.team2412.robot;
 
 public class Hardware {
-	// this is where we will ID the hardware, like motors n stuff
+	// apriltags stuff
+	public static final String PHOTON_CAM = "Arducam_OV9281_USB_Camera";
+	public static final String PHOTON_IP = "10.24.12.51";
```
jamesdooley4: This needs to be between 10.TE.AM.5 (or .6, if the driver station is statically set to .5) to 10.TE.AM.20 (inclusive) per https://docs.wpilib.org/en/stable/docs/networking/networking-introduction/ip-configurations.html#on-the-field-static-configuration
Anything above .20 might be allocated elsewhere by the FMS.

src/main/java/frc/team2412/robot/util/DrivebaseWrapper.java
```diff
@@ -0,0 +1,99 @@
...
+	// SwerveDrive's methods (which are thread-safe)
+	private final VisionMeasurementAdder visionMeasurementAdder;
+
+	/** Constructor for when there is no drivebase. */
+	public DrivebaseWrapper() {
+		field = new Field2d();
+		// YAGSL will normally do this for us
+		// This may be pointless though since Field2d can't be shown in both SmartDashboard and the
+		// MatchDashboard tab
+		SmartDashboard.putData("Field", field);
```
TAKBS2412, line 41: Should update this name to specifty what measurements this field object is using (combined vision + drivebase).

src/main/java/frc/team2412/robot/util/MatchDashboard.java
```diff
@@ -15,7 +15,7 @@ public class MatchDashboard {
 	private final Field2d field;
 
 	public MatchDashboard(Subsystems s) {
-		field = s.drivebaseSubsystem.getField();
+		field = s.drivebaseWrapper.getField();
```
TAKBS2412, line 18: Since these field objects track two separate things (odometry only pose and vision + odometry pose), we should probably log them separately, and label them as such.

TAKBS2412: (and if they're already logged separately, make it clear which one is using which measurements, like for `DrivebaseWrapper`).

src/main/java/frc/team2412/robot/Hardware.java
TAKBS2412: We should probably update this IP to be within the range that the docs suggest: https://docs.wpilib.org/en/stable/docs/networking/networking-introduction/ip-configurations.html#on-the-field-static-configuration

jamesdooley4 (dismissed):

TAKBS2412 (dismissed):

src/main/java/frc/team2412/robot/util/DrivebaseWrapper.java
jamesdooley4: And per the comment here, this is logging the same field object here and in `MatchDashboard`, which seems potentially wasteful in bandwidth & processing.

src/main/java/frc/team2412/robot/util/MatchDashboard.java
jamesdooley4: I think they log the same thing (both end up logging the `field` class member of `DrivebaseWrapper`

vendordeps/photonlib.json
```diff
@@ -0,0 +1,57 @@
+{
+    "fileName": "photonlib.json",
+    "name": "photonlib",
+    "version": "v2024.2.4",
```
jamesdooley4, line 4: FYI that latest version is v2024.2.6

jamesdooley4 (dismissed):

src/main/java/frc/team2412/robot/util/MatchDashboard.java
KangarooKoala: Yeah, there should just be one `Field2d` instance (with different `FieldObject2d` instances with appropriate names).

src/main/java/frc/team2412/robot/util/DrivebaseWrapper.java
KangarooKoala: The intent was to make the robot code behave as similarly as possible with and without drivebase, and YAGSL will always put the field to SmartDashboard (which incidentally seems to be the only spot the sim GUI will render the field). It doesn't waste anything in bandwidth or processing, since the [`Field2d`](https://github.com/wpilibsuite/allwpilib/blob/main/wpilibj/src/main/java/edu/wpi/first/wpilibj/smartdashboard/Field2d.java) implementation only keeps track of the last spot it was put. (But it is weird to have it in two different places)

jamesdooley4: Agree that it seems weird, but doesn't seem like a big problem so is fine to fix later.

vendordeps/photonlib.json
KangarooKoala: Looks like [v2024.2.5](https://github.com/PhotonVision/photonvision/releases/tag/v2024.2.5) and [v2024.2.6](https://github.com/PhotonVision/photonvision/releases/tag/v2024.2.6) are very new, and v2024.2.6 has substantial changes that will take time to investigate (without a lot of added functionality). We'll probably do it later in the season, but I'd like to get this PR merged relatively soon.

src/main/java/frc/team2412/robot/sensors/AprilTagsProcessor.java
```diff
@@ -0,0 +1,150 @@
...
+ */
+public class AprilTagsProcessor {
+	public static final Transform3d ROBOT_TO_CAM =
+			new Transform3d(
+					Units.inchesToMeters(27.0 / 2.0 - 0.94996),
+					0,
+					Units.inchesToMeters(8.12331),
+					new Rotation3d(0, Units.degreesToRadians(-30), 0));
+
+	// TODO Measure these
```
bobtseattle, line 51: Given the time it would seem to take to get a statistically significant number of measurements to accurately measure the "jitter" or spread of the measurements of the camera (My recollection is that the literature recommends hundreds to thousands of measurements when calculating an accurate standard deviation of a sensor for a kalman filter.), we might want to keep them around the range as they are now, just like jack and many other teams do.

Seemingly, we will want to err on the side of std devs too large or the kalman filter will put too much faith in the accuracy of the AprilTags vs. our known position and IMU measurements which means an inaccurate measurement from the AprilTags at the edge of the range could throw everything off.

bobtseattle (approved):
Just one note on calculating statistically significant std deviations for the kalman/bayesian filters and given these numbers aren't too far off what others use in FRC.

https://github.com/FRCTeam2910/2023CompetitionRobot-Public/blob/7f38faa9afbd41e4372ae7e518dbff692a5abdeb/src/main/java/org/frcteam2910/c2023/subsystems/drivetrain/DrivetrainSubsystem.java#L240-L241

TAKBS2412 (approved):

src/main/java/frc/team2412/robot/sensors/AprilTagsProcessor.java
KangarooKoala: Alright, we'll keep this in mind for later.

PR #43, opened 2024-02-16
'

SMKettle:


jamesdooley4:
Please add a description...

src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -50,13 +57,13 @@ public Controls(Subsystems s) {
...
-		driveIntakeInButton = driveController.x();
-		driveIntakeStopButton = driveController.b();
-		driveIntakeReverseButton = driveController.y();
-		codriveIntakeInButton = codriveController.povLeft();
-		codriveIntakeStopButton = codriveController.povRight();
-		codriveIntakeReverseButton = codriveController.povUp();
+		// intake buttons (may change later)
+		driveIntakeInButton = codriveController.povLeft();
+		driveIntakeStopButton = codriveController.povRight();
+		driveIntakeSpitButton = codriveController.povUp();
```
jamesdooley4, line 63: Drive buttons can't be on the codrive controller. Either this is using the wrong controller, or the button variable name is wrong.

jamesdooley4: These drive buttons also aren't mapped to commands anywhere, so don't actually do anything...

jamesdooley4 (changes requested):

src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -106,4 +102,26 @@ private void bindLauncherControls() {
...
+		codriveIntakeStopButton.onTrue(
+				Commands.parallel(
+						new IntakeStopCommand(s.intakeSubsystem),
+						new IndexStopCommand(s.intakeSubsystem),
+						new FeederStopCommand(s.intakeSubsystem)));
+		codriveIntakeSpitButton.onTrue(
+				Commands.parallel(
+						new IntakeReverseCommand(s.intakeSubsystem),
+						new IndexReverseCommand(s.intakeSubsystem),
+						new FeederReverseCommand(s.intakeSubsystem)));
```
jamesdooley4, line 125: These will all crash since there's this check in the ParallelCommandGroup each of these create:
From https://github.com/wpilibsuite/allwpilib/blob/6afff996401343c3bfcc46a40a40abda429e5077/wpilibNewCommands/src/main/java/edu/wpi/first/wpilibj2/command/ParallelCommandGroup.java#L51C1-L54C8
```
      if (!Collections.disjoint(command.getRequirements(), m_requirements)) {
        throw new IllegalArgumentException(
            "Multiple commands in a parallel composition cannot require the same subsystems");
      }
```

jamesdooley4: For these you likely need to add a new command that does the combination that's needed. And to address the proliferation of tiny command classes that just call a method or two on a subsystem, consider adding command helpers to the subsystem (like DrivebaseSubsystem::rotateToAngle did) rather than small Command classes.

Issue #44, opened 2024-02-16
SetAngleLaunchCommand should be reworked to allow angle and speed to be set in parallel


jbko6:
See https://github.com/robototes/Crescendo2024/pull/40#discussion_r1491915840

iamawesomecat:
So if i understand this correctly, you want me to change it from sequential to parallel?

PR #49, opened 2024-02-17
Launcher v1.2

iamawesomecat:
Added a manual control for the pivot. 
         This was done by reworking the setAngleCommand to make it intake a speed value, and made a method in the subsystem 
              to compliment that.

Additionally, fixed issue #44
                 Just set the command to be a normal command, and gave it the responsibilities of the old setAngle, just with the added responsibility of setting flywheel speed.
                 
Added more logging stuff


TAKBS2412 (approved):
Thanks for the detailed PR description!

PR #50, opened 2024-02-17
Intake subsystem

eddiert1239:
added 3rd sensor
changed regular in commands to just return true for isfinished
when using all in command, the bottom intake motors stop after the indexsensor sees


eddiert1239:
this is replacing the " ' " pr

TAKBS2412 (commented):
Curious, what's the third sensor used for? I don't see it being used anywhere.

TAKBS2412 (commented):
LGTM pending the response to my previous question.

eddiert1239:
> Curious, what's the third sensor used for? I don't see it being used anywhere.

It isn't being used right now, I don't know what it would be used for but it is apparently on the robot. Should we remove it or implement it or what?

TAKBS2412:
> > Curious, what's the third sensor used for? I don't see it being used anywhere.
> 
> It isn't being used right now, I don't know what it would be used for but it is apparently on the robot. Should we remove it or implement it or what?

If you're not sure what it would be used for, I'd say remove it, as it's not too hard to add it back later if we need to.

TAKBS2412 (approved):

PR #51, opened 2024-02-17
Enable FOC control for drivebase

jbko6:
Need to actually license the motors before we can use this. We bought the season pass earlier this year so we can license up to 60 devices including 2 CANivores which can license their connected motors. I'll probably try and license Bonk, the practice bot, and the comp bot.

Might make sense to add an easier way to disable FOC in code in the future.


jamesdooley4 (approved):

jbko6:
Waiting to merge until we can test this and decide if we actually want FOC

TAKBS2412 (changes requested):
Waiting until post glacier peak, drive team preferred non FOC for now.

PR #52, opened 2024-02-19
Coast motors on disable

jbko6:


src/main/java/frc/team2412/robot/Robot.java
```diff
@@ -114,11 +116,15 @@ public void autonomousInit() {
 		DriverStation.silenceJoystickConnectionWarning(!DriverStation.isFMSAttached());
 
 		autoChooser.getSelected().schedule();
+
+		subsystems.drivebaseSubsystem.setMotorBrake(true);
```
KangarooKoala: Would it be better to put this (and the one in `teleopInit()`) in `disabledExit()`? (Note that `testInit()` wasn't updated, which is probably fine since we don't use it very often, but it could've caused some confusion later on)

jamesdooley4 (dismissed):

jamesdooley4 (approved):

PR #59, opened 2024-02-21
Led subsystem

Jetblackdragon:


jamesdooley4 (dismissed):

jamesdooley4:
Need to fix spotless errors...

jamesdooley4 (approved):

kirbt:
Do we wanna also include commands to set tbe LEDS?

jamesdooley4:
> Do we wanna also include commands to set tbe LEDS?

That can be in the next PR. I don't think we need to block this PR on adding those.

PR #60, opened 2024-02-21
launcher 1.3

iamawesomecat:
Added variation to testinit depending on the enabled subsystems, uses that to run specific diagnostics.

Added a autoAlign utility function for trap. also contains a command

Logged flywheel temps

Fixed some motor issues in sim

plus some fun spotless commits :)


src/main/java/frc/team2412/robot/Robot.java
```diff
@@ -101,10 +104,20 @@ public void robotInit() {
...
-		CommandScheduler.getInstance()
-				.schedule(
-						new diagnosticSequentialCommand(
-								subsystems.intakeSubsystem, subsystems.launcherSubsystem));
+		if (SubsystemConstants.INTAKE_ENABLED && SubsystemConstants.LAUNCHER_ENABLED) {
+			CommandScheduler.getInstance()
+					.schedule(
+							new diagnosticSequentialCommand(
+									subsystems.intakeSubsystem, subsystems.launcherSubsystem));
+		} else if (SubsystemConstants.INTAKE_ENABLED) {
```
bobtseattle: What is the use case for these being if else?  This requires subsystems to be disabled / enabled and then a full recompile to run diagnostic commands for different subsystems. 
 Also there doesn't seem to be a way to run diagnostics for multiple systems since only one condition can be satisfied.

bobtseattle (changes requested):
While it will be good to only run diagnostics for enabled systems, the logic for this requires recompiling to run different subsystems.

src/main/java/frc/team2412/robot/Robot.java
iamawesomecat: Its so that only the enabled subsystems are tested (it didnt work with them being disabled, because of addrequrirements), this makes it so only the necessary subsystems are tested. Making testing easy if wanna work on a subsystem by subsystem basis.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -76,63 +80,101 @@ public class LauncherSubsystem extends SubsystemBase {
...
-		launcherBottomMotor.setSmartCurrentLimit(20);
-		launcherAngleMotor.setSmartCurrentLimit(20);
+		launcherTopMotor.setSmartCurrentLimit(40);
+		launcherBottomMotor.setSmartCurrentLimit(40);
+		launcherAngleOneMotor.setSmartCurrentLimit(20);
+		launcherAngleTwoMotor.setSmartCurrentLimit(20);
 
-		launcherAngleMotor.setSoftLimit(CANSparkBase.SoftLimitDirection.kForward, 100);
-		launcherAngleMotor.setSoftLimit(CANSparkBase.SoftLimitDirection.kReverse, 2);
+		launcherAngleOneMotor.setSoftLimit(CANSparkBase.SoftLimitDirection.kForward, 100);
```
bobtseattle, line 159: We have to be careful when first testing these limits so we do not break anything.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -76,63 +80,101 @@ public class LauncherSubsystem extends SubsystemBase {
...
 		// inveritng the bottom motor lmao
 		launcherBottomMotor.setInverted(true);
 
 		// current limit
-		launcherTopMotor.setSmartCurrentLimit(20);
-		launcherBottomMotor.setSmartCurrentLimit(20);
-		launcherAngleMotor.setSmartCurrentLimit(20);
+		launcherTopMotor.setSmartCurrentLimit(40);
+		launcherBottomMotor.setSmartCurrentLimit(40);
+		launcherAngleOneMotor.setSmartCurrentLimit(20);
```
bobtseattle, line 156: We should disable one of the angle motors when first testing the pivot so the motors do not fight each other and burn one out.

bobtseattle (commented):
Just a couple of notes we should be aware of when testing.

src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -75,14 +77,23 @@ public Controls(Subsystems s) {
...
-					.leftBumper()
+					.leftTrigger()
 					.whileTrue(
 							new FullTargetCommand(
 									s.launcherSubsystem,
 									s.intakeSubsystem,
 									s.drivebaseSubsystem,
 									this,
-									driveController.rightBumper()));
+									driveController.leftTrigger()));
```
jbko6: If we're switching to triggers (which I don't suggest), this needs to be changed to rightTrigger or some other button, otherwise FullTargetCommand will launch as soon as it begins. This parameter describes the button that, when pressed, will launch the note

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
```diff
@@ -58,6 +55,11 @@ public FullTargetCommand(
 	public void initialize() {
 		CommandScheduler.getInstance().schedule(yawAlignmentCommand);
 		intakeSubsystem.feederStop();
+
+		SPEAKER_POSE =
```
jbko6, line 59: For those curious about this change, this fix was required to resolve a crashing bug in simulation. Because the Alliance is reported as INVALID on sim startup, getting the alliance would crash in sim. This issue is only observed in simulation which is why it wasn't caught in its original PR.

src/main/java/frc/team2412/robot/Controls.java
iamawesomecat: if you want i can switch it to bumpers, or ask drive team for a preference

iamawesomecat: currently waiting on drive team's response

iamawesomecat: Driveteam was fine with it, changed to bumper.

src/main/java/frc/team2412/robot/util/TrapAlign.java
```diff
@@ -0,0 +1,102 @@
...
+	private static class AlignCommand extends Command {
+		private final DrivebaseSubsystem drivebaseSubsystem;
+		private final LauncherSubsystem launcherSubsystem;
+		private Command trapCommand = null;
+
+		public AlignCommand(
+				DrivebaseSubsystem drivebaseSubsystem, LauncherSubsystem launcherSubsystem) {
+			this.drivebaseSubsystem = drivebaseSubsystem;
+			this.launcherSubsystem = launcherSubsystem;
+			addRequirements(launcherSubsystem);
```
bobtseattle: Does drivebaseSubsystem need to be part of the requirements?

bobtseattle (changes requested):
Need to make sure all subsystems are in requirements so commands do not conflict.

src/main/java/frc/team2412/robot/util/TrapAlign.java
iamawesomecat: yep. my bad, fixed it tho

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -76,63 +82,101 @@ public class LauncherSubsystem extends SubsystemBase {
...
+		launcherAngleTwoMotor.setSoftLimit(CANSparkBase.SoftLimitDirection.kForward, 100);
+		launcherAngleTwoMotor.setSoftLimit(CANSparkBase.SoftLimitDirection.kReverse, 25);
 
 		launcherTopMotor.burnFlash();
 		launcherBottomMotor.burnFlash();
-		launcherAngleMotor.burnFlash();
+		launcherAngleOneMotor.burnFlash();
+		launcherAngleTwoMotor.burnFlash();
+
+		launcherAngleTwoMotor.follow(launcherAngleOneMotor);
```
TAKBS2412: Should follow before burning flash so that `burnFlash()` saves the follower setting.

iamawesomecat: 👍

src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -120,18 +132,19 @@ private void bindLauncherControls() {
 						s.launcherSubsystem,
 						new SetAngleCommand(
 								s.launcherSubsystem,
-								() -> MathUtil.applyDeadband(codriveController.getLeftY(), 0.1) * 0.75));
+								() -> MathUtil.applyDeadband(codriveController.getLeftY(), 0.1) * 0.1));
```
TAKBS2412, line 135: Note that this may be too low, as the max speed won't exceed 0.1. Should be okay to leave as-is though just to be safe.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -143,11 +188,16 @@ public void configMotors() {
 		launcherBottomPIDController.setI(0);
 		launcherBottomPIDController.setD(0);
 		launcherBottomPIDController.setFF(0);
+
+		launcherAngleOneMotor.getEncoder().setPosition(launcherAngleEncoder.getPosition());
+		launcherAngleOneMotor.getEncoder().setPositionConversionFactor(80);
```
TAKBS2412: Did we check with mech to make sure 80 is the correct ratio? From my memory it was 180, but could be misremembering.

iamawesomecat: jonah told me 80, but ill go ask

bobtseattle: Would also be good to use constants so we don't forget one if we need to update.

iamawesomecat: Yeah they said 180, whoopsies

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -76,63 +82,102 @@ public class LauncherSubsystem extends SubsystemBase {
...
+		launcherAngleOneMotor.restoreFactoryDefaults();
+		launcherAngleTwoMotor.restoreFactoryDefaults();
 		// idle mode (wow)
 		launcherTopMotor.setIdleMode(IdleMode.kCoast);
 		launcherBottomMotor.setIdleMode(IdleMode.kCoast);
-		launcherAngleMotor.setIdleMode(IdleMode.kBrake);
+		launcherAngleOneMotor.setIdleMode(IdleMode.kBrake);
+		launcherAngleTwoMotor.setIdleMode(IdleMode.kBrake);
 		// inveritng the bottom motor lmao
 		launcherBottomMotor.setInverted(true);
```
bobtseattle, line 150: Now that we see the motors installed, it does appear we need to invert one of the angle motors so they do not work against each other and break pivot.  The one piece they told us not to break.

bobtseattle (changes requested):
I would recommend getting away from hardcoding values and starting to use constants.

src/main/java/frc/team2412/robot/Controls.java
iamawesomecat: Yeah, i just wanted to put something i knew would be safe and wouldnt like destroy everything lol. ill turn it up to .3 tho

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: Fixed.

bobtseattle (approved):
LGTM

PR #63, opened 2024-02-22
Add display of selected autonomous path

KangarooKoala:
Hopefully this will help with learning the system behind the auto names.

I included methods to pre-load everything in case there were performance concerns, but they probably won't be necessary (especially since we shouldn't be changing the selected auto during matches).

### DIFF
--- a/src/main/java/frc/team2412/robot/util/PathPlannerAutos.java
+++ b/src/main/java/frc/team2412/robot/util/PathPlannerAutos.java
@@ -1,5 +1,7 @@
 package frc.team2412.robot.util;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.pathplanner.lib.path.PathPlannerPath;
 import com.pathplanner.lib.path.PathPlannerTrajectory;
 import edu.wpi.first.math.geometry.Rotation2d;
@@ -8,7 +10,6 @@ import edu.wpi.first.wpilibj.DriverStation;
 import edu.wpi.first.wpilibj.Filesystem;
 import java.io.File;
 import java.io.IOException;
-import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


jbko6 (dismissed):

src/main/java/frc/team2412/robot/util/PathPlannerAutos.java
```diff
@@ -0,0 +1,247 @@
...
+		return text.substring(start, Math.min(comma, objectEnd));
+	}
+
+	private static boolean readJsonBoolean(String text, int start) {
+		// False for values that are not "true" or "false" (case insensitive)
+		return Boolean.parseBoolean(readJsonValue(text, start));
+	}
+
+	private static double readJsonDouble(String text, int start) throws NumberFormatException {
+		return Double.parseDouble(readJsonValue(text, start));
```
jamesdooley4: Consider using one of the Json libraries wpilib uses internally for parsing Json instead of writing your own simple parser: https://github.com/FasterXML/jackson or https://github.com/google/gson
Not blocking for this PR, but would help make this more resilient and ideally re-use code already linked into the program.

jamesdooley4 (dismissed):

src/main/java/frc/team2412/robot/util/PathPlannerAutos.java
KangarooKoala: Oops, I was worried about adding a dependency but forgot WPILib already uses some JSON libraries. I'll switch to using FasterXML/jackson tomorrow.

KangarooKoala:
It looks like loading might be a little bit slower (as in not instant), but it should be fine- It was just a little stutter that I'm not even 100% sure was caused by this change.

KangarooKoala:
Regarding the TODO for the named commands, there's work in progress on conditional paths that would be implemented through those named commands, but I think I'll add the logic for that after that gets merged in. It isn't high importance, so I don't think this PR should be blocked on that.

TAKBS2412 (approved):

jamesdooley4 (approved):

PR #64, opened 2024-02-23
[Low priority] Separate formatting and automated test CI runs

KangarooKoala:
I dropped the check for changes added by 0c75603 since I couldn't reproduce a case where `spotlessCheck` would silently make changes, but I could add that back just to be sure.


jamesdooley4 (approved):

bobtseattle (approved):
LGTM

TAKBS2412 (approved):

PR #68, opened 2024-02-24
Intake and Sensors

Yumnah2:


src/main/java/frc/team2412/robot/commands/intake/AllInCommand.java
```diff
@@ -34,6 +34,9 @@ public void end(boolean interrupted) {
 
 	@Override
 	public boolean isFinished() {
-		return intakeSubsystem.getFeederSensor();
+		if (!intakeSubsystem.getSensorOverride()) {
```
TAKBS2412: This can be rewritten to use ||.

eddiert1239: yeah you're right

src/main/java/frc/team2412/robot/commands/intake/AllInCommand.java
```diff
@@ -34,6 +34,6 @@ public void end(boolean interrupted) {
 
 	@Override
 	public boolean isFinished() {
-		return intakeSubsystem.getFeederSensor();
+		return !intakeSubsystem.getSensorOverride() || intakeSubsystem.getFeederSensor();
```
TAKBS2412: I think this is different from the older version of the logic, if sensor override is true but feeder sensor is false, the old version would return true but the new version returns false.

eddiert1239: yeah you're right

src/main/java/frc/team2412/robot/commands/intake/AllInCommand.java
```diff
@@ -20,7 +20,7 @@ public void initialize() {
 
 	@Override
 	public void execute() {
-		if (intakeSubsystem.getIndexSensor()) {
+		if (intakeSubsystem.getIndexSensor() && !intakeSubsystem.getSensorOverride()) {
```
bobtseattle: To make the logic more clear, is it better to override the sensor here or in the getSensorOverride method?
Or with a parameter to getSensorOverride that overrides the sensor?

bobtseattle (commented):
I am debating if there is a better way to override the sensor value so we don't have to implement extra logic in this and other commands(if ever needed)?

eddiert1239:
> I am debating if there is a better way to override the sensor value so we don't have to implement extra logic in this and other commands(if ever needed)?

Both the getFeederSensor() and getIndexSensor() methods now return sensor value if not overriding, and false if we are overriding. This means no extra logic in commands.

bobtseattle (commented):
LGTM.  Would be nice if the sensor method would return true to make it more intuitive, but we can go with the ! until we get confused.

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
```diff
@@ -214,12 +223,16 @@ public void feederShoot() {
 	}
 
 	// sensor methods
-	public boolean getIndexSensor() {
-		return indexSensor.get();
+	public boolean indexSensorHasNote() {
+		return !indexSensor.get() && !getSensorOverride();
```
KangarooKoala, line 227: Why is `indexSensor.get()` inverted now? (This also applies to `feederSensorHasNote()`)

KangarooKoala (commented):
Looks pretty good to me, except the one question (and also the correct spelling is ingest, not injest).

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
eddiert1239: .get() returns true as a default, and if it senses something it returns false

TAKBS2412 (approved):

jamesdooley4 (approved):

PR #71, opened 2024-02-27
Rumble to Intake

Yumnah2:


src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -121,11 +124,15 @@ private void bindDrivebaseControls() {
...
+		driveIntakeInButton.onTrue(
+				Commands.sequence(
+						new AllInCommand(s.intakeSubsystem, this),
+						new WaitCommand(0.5),
+						new StopRumbleCommand(this)));
 		driveIntakeStopButton.onTrue(new AllStopCommand(s.intakeSubsystem));
 		driveIntakeReverseButton.onTrue(new AllReverseCommand(s.intakeSubsystem));
 		driveIntakeRejectButton.onTrue(new IntakeRejectCommand(s.intakeSubsystem));
-		codriveIntakeInButton.onTrue(new AllInCommand(s.intakeSubsystem));
+		codriveIntakeInButton.onTrue(new AllInCommand(s.intakeSubsystem, this));
```
TAKBS2412, line 139: Curious why this is different from the drive intake in button?

src/main/java/frc/team2412/robot/util/AutoLogic.java
```diff
@@ -65,7 +65,7 @@ public void registerCommands() {
 						.andThen(new WaitCommand(1))
 						.andThen(new FeederInCommand(s.intakeSubsystem)));
 
-		NamedCommands.registerCommand("Intake", new AllInCommand(s.intakeSubsystem));
+		NamedCommands.registerCommand("Intake", new AllInCommand(s.intakeSubsystem, controls));
```
TAKBS2412: Note that since this is in auto we don't need to rumble the drive controller... probably okay to leave in though.

src/main/java/frc/team2412/robot/commands/intake/AllInCommand.java
```diff
@@ -30,6 +33,7 @@ public void end(boolean interrupted) {
 		intakeSubsystem.intakeReject();
 		intakeSubsystem.indexStop();
 		intakeSubsystem.feederStop();
+		controls.vibrateDriveController(1.0);
```
jamesdooley4: Since the command can be interrupted for reasons other than `isFinished()` being signaled (could be aborted if scheduled in parallel with a wait command, a different button is pressed that starts a different command using the intake subsystem, etc.), it'd be best to check and only rumble the controller if `intakeSubsystem.feederSensorHasNote()` is true.

jamesdooley4: Or, more simply, only vibrate the controller if the `interrupted` parameter is `false` (meaning the command is ending because it finished and wasn't interrupted).

src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -121,11 +124,15 @@ private void bindDrivebaseControls() {
 	private void bindIntakeControls() {
 		// CommandScheduler.getInstance()
 		// 		.setDefaultCommand(s.intakeSubsystem, new IntakeStopCommand(s.intakeSubsystem));
-		driveIntakeInButton.onTrue(new AllInCommand(s.intakeSubsystem));
+		driveIntakeInButton.onTrue(
+				Commands.sequence(
+						new AllInCommand(s.intakeSubsystem, this),
+						new WaitCommand(0.5),
+						new StopRumbleCommand(this)));
```
jamesdooley4: I'm pretty sure the controller would get stuck vibrating if the driver or codriver pressed any of the other intake buttons right after a note is detected & the controller starts rumbling. That would interrupt this sequence during the WaitCommand so the StopRumbleCommand would never run. One way to fix that is to wrap AllInCommand and WaitCommand in a new command class derived from SequentialCommandGroup and have the `end(..)` method call controls.vibrateDriveController(0.0).

src/main/java/frc/team2412/robot/util/AutoLogic.java
jamesdooley4: `AllInCommand` could be updated to tolerate `null` being passed for `controls` and skip the controller vibration in that case. That'd avoid the controllers vibrating during auto and potentially falling off the drive board.

jamesdooley4 (changes requested):

src/main/java/frc/team2412/robot/commands/intake/AllInRumbleCommand.java
```diff
@@ -0,0 +1,23 @@
...
+import frc.team2412.robot.subsystems.IntakeSubsystem;
+
+public class AllInRumbleCommand extends SequentialCommandGroup {
+	private final IntakeSubsystem intakeSubsystem;
+	private final Controls controls;
+
+	public AllInRumbleCommand(IntakeSubsystem intakeSubsystem, Controls controls) {
+		this.intakeSubsystem = intakeSubsystem;
+		this.controls = controls;
+		addRequirements(intakeSubsystem);
```
jamesdooley4: This implicitly happens in the `addCommands` call below (see https://github.com/wpilibsuite/allwpilib/blob/f1a1ffd7fc96649df2293e6337e02f25b557682c/wpilibNewCommands/src/main/java/edu/wpi/first/wpilibj2/command/SequentialCommandGroup.java#L51 ) so should be removed here

src/main/java/frc/team2412/robot/commands/intake/AllInRumbleCommand.java
```diff
@@ -0,0 +1,22 @@
...
+	private final Controls controls;
+
+	public AllInRumbleCommand(IntakeSubsystem intakeSubsystem, Controls controls) {
+		this.intakeSubsystem = intakeSubsystem;
+		this.controls = controls;
+
+		addCommands(new AllInCommand(intakeSubsystem, controls), new WaitCommand(0.5));
+	}
+
+	public void end() {
```
TAKBS2412: Note that we will want to call the superclass's `end()` method to make sure the command group responds to cancellations properly.

src/main/java/frc/team2412/robot/commands/intake/AllInCommand.java
```diff
@@ -30,6 +33,10 @@ public void end(boolean interrupted) {
 		intakeSubsystem.intakeReject();
 		intakeSubsystem.indexStop();
 		intakeSubsystem.feederStop();
+
+		if (controls != null && interrupted == false) {
```
TAKBS2412: (Nit: you can use `!interrupted` instead of directly comparing against `false`.)

iamawesomecat:
We should prob get this merged today

src/main/java/frc/team2412/robot/commands/intake/AllInCommand.java
```diff
@@ -20,16 +25,66 @@ public void initialize() {
...
+			intakeSubsystem.intakeBackReject();
+			intakeSubsystem.intakeLeftReject();
+			intakeSubsystem.intakeRightReject();
+
+			if (controls != null) {
+				Commands.race(new RumbleCommand(controls), new WaitCommand(0.5)).schedule();
+			}
+		}
+
+		if (intakeSubsystem.intakeBackSeesNote()) {
```
jamesdooley4: What should happen if more than one side sees a note?

src/main/java/frc/team2412/robot/commands/intake/AllInCommand.java
```diff
@@ -20,16 +25,66 @@ public void initialize() {
...
+			if (controls != null) {
+				Commands.race(new RumbleCommand(controls), new WaitCommand(1)).schedule();
+			}
 		}
 	}
 
 	@Override
 	public void end(boolean interrupted) {
-		// intakeSubsystem.intakeReject();
+		intakeSubsystem.intakeReject();
```
jamesdooley4: Since stop is called on the next line, the intake motors won't have time to reverse, and will just stop. Recommend deleting the call to intakeReject.

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
```diff
@@ -140,6 +147,11 @@ public IntakeSubsystem() {
 		indexSensor = new DigitalInput(INDEX_SENSOR);
 		feederSensor = new DigitalInput(FEEDER_SENSOR);
 
+		intakeFrontSensor = intakeMotorFront.getForwardLimitSwitch(SparkLimitSwitch.Type.kNormallyOpen);
+		intakeBackSensor = intakeMotorBack.getForwardLimitSwitch(SparkLimitSwitch.Type.kNormallyOpen);
+		intakeLeftSensor = intakeMotorLeft.getForwardLimitSwitch(SparkLimitSwitch.Type.kNormallyOpen);
+		intakeRightSensor = intakeMotorRight.getForwardLimitSwitch(SparkLimitSwitch.Type.kNormallyOpen);
```
jamesdooley4, line 92: Also need to update the motor configuration code so the controller knows the limit switch is plugged in (the same wire can also be used for an encoder), and to not stop the motor when the switch is signaled

jamesdooley4 (changes requested):

src/main/java/frc/team2412/robot/commands/intake/AllInCommand.java
Yumnah2: We talked to Dom and he said that one sensor would probably sense a note before another one, and therefore the second sensor and intake roller would start rejecting the second note. But we should probably test it on the robot to be sure

src/main/java/frc/team2412/robot/commands/intake/AllInCommand.java
TAKBS2412: I think this is intake reject followed by index stop, so the intake motors will keep rejecting.

src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -122,11 +122,12 @@ private void bindDrivebaseControls() {
 	private void bindIntakeControls() {
 		// CommandScheduler.getInstance()
 		// 		.setDefaultCommand(s.intakeSubsystem, new IntakeStopCommand(s.intakeSubsystem));
-		driveIntakeInButton.onTrue(new AllInCommand(s.intakeSubsystem));
-		// driveIntakeStopButton.onTrue(new AllStopCommand(s.intakeSubsystem));
+		driveIntakeInButton.onTrue(new RumbleCommand(this));
```
TAKBS2412: Should this be `AllInCommand`?

src/main/java/frc/team2412/robot/Subsystems.java
```diff
@@ -14,8 +14,8 @@ public static class SubsystemConstants {
 		private static final boolean IS_COMP = Robot.getInstance().isCompetition();
 
 		public static final boolean APRILTAGS_ENABLED = true;
-		public static final boolean LIMELIGHT_ENABLED = false;
-		public static final boolean CLIMB_ENABLED = false;
+		public static final boolean LIMELIGHT_ENABLED = true;
+		public static final boolean CLIMB_ENABLED = true;
```
TAKBS2412: These should be reverted to `false`.

src/main/java/frc/team2412/robot/commands/intake/AllInCommand.java
```diff
@@ -20,16 +25,98 @@ public void initialize() {
...
+				intakeSubsystem.intakeLeftReject();
+				intakeSubsystem.intakeRightReject();
+			} else {
+				intakeSubsystem.intakeBackStop();
+				intakeSubsystem.intakeLeftStop();
+				intakeSubsystem.intakeRightStop();
+			}
+
+			if (controls != null) {
+				Commands.race(new RumbleCommand(controls), new WaitCommand(1)).schedule();
```
TAKBS2412: FYI that you can use [`withTimeout()`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/Command.html#withTimeout(double)) (not required, but useful tip).

src/main/java/frc/team2412/robot/commands/intake/AllInCommand.java
```diff
@@ -20,16 +25,98 @@ public void initialize() {
...
 		if (intakeSubsystem.indexSensorHasNote()) {
-			intakeSubsystem.intakeReject();
+			if (!intakeSubsystem.getRejectOverride()) {
+				intakeSubsystem.intakeReject();
+			} else {
+				intakeSubsystem.intakeStop();
+			}
+
+			if (controls != null) {
+				Commands.race(new RumbleCommand(controls), new WaitCommand(3)).schedule();
```
TAKBS2412, line 95: Is there a reason why the rumble here is for 3 seconds, but for back, left, and right it's 2 seconds, and only 1 second for the front? (maybe a method of signaling to the drive team?)

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
```diff
@@ -47,8 +48,13 @@ public class IntakeSubsystem extends SubsystemBase {
 	private final DigitalInput indexSensor;
 	private final DigitalInput feederSensor;
 
+	private final SparkLimitSwitch intakeFrontSensor;
+	private final SparkLimitSwitch intakeBackSensor;
+	private final SparkLimitSwitch intakeLeftSensor;
+	private final SparkLimitSwitch intakeRightSensor;
```
TAKBS2412: We should probably log these to help debug any issues in the future.

src/main/java/frc/team2412/robot/commands/intake/AllInCommand.java
```diff
@@ -20,16 +25,98 @@ public void initialize() {
...
+				intakeSubsystem.intakeLeftReject();
+				intakeSubsystem.intakeRightReject();
+			} else {
+				intakeSubsystem.intakeFrontStop();
+				intakeSubsystem.intakeLeftStop();
+				intakeSubsystem.intakeRightStop();
+			}
+
+			if (controls != null) {
+				Commands.race(new RumbleCommand(controls), new WaitCommand(2)).schedule();
```
TAKBS2412, line 45: Note that this will cause a new rumble command to be scheduled every ~20 ms until the note stops triggering the sensor, which would be a bit excessive (and might lead to a crash depending on the maximum number of commands).

TAKBS2412 (changes requested):

src/main/java/frc/team2412/robot/commands/intake/AllInCommand.java
Yumnah2: Sorry I forgot to change the intake front. The intake stops/rejects should be 2 seconds, index stopping/rejecting all of the intake should be 3, and end() should be 3 as well.

src/main/java/frc/team2412/robot/Subsystems.java
Yumnah2: fixed, thanks

src/main/java/frc/team2412/robot/Controls.java
Yumnah2: while testing we changed it to see if rumble worked on the controllers so we'll change it back to AllInCommand soon

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
```diff
@@ -140,11 +155,17 @@ public IntakeSubsystem() {
...
+		intakeBackSensor = intakeMotorBack.getForwardLimitSwitch(SparkLimitSwitch.Type.kNormallyOpen);
+		intakeLeftSensor = intakeMotorLeft.getForwardLimitSwitch(SparkLimitSwitch.Type.kNormallyOpen);
+		intakeRightSensor = intakeMotorRight.getForwardLimitSwitch(SparkLimitSwitch.Type.kNormallyOpen);
+
 		resetMotors();
 
 		ShuffleboardTab shuffleboardTab = Shuffleboard.getTab("Intake");
 		shuffleboardTab.addBoolean("Index Sensor - ", this::indexSensorHasNote).withSize(1, 1);
 		shuffleboardTab.addBoolean("Feeder Sensor - ", this::feederSensorHasNote).withSize(1, 1);
+		shuffleboardTab.addBoolean("Intake Sensor - ", this::intakeBackSeesNote).withSize(1, 1);
```
bobtseattle: Should this be back or if any intake sensors are triggered?

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
Yumnah2: added them to shuffleboard

src/main/java/frc/team2412/robot/commands/intake/AllInCommand.java
Yumnah2: This should be fixed now

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
Yumnah2: We updated the configuration but not sure if it'll impact the motor

TAKBS2412 (dismissed):
Approving to have a baseline before robot testing.

Note that the rumbled variables like `rumbledIntakeFront` aren't reset back to `false` once the command ends, so the controller will only rumble the first time the robot intakes a note on a given side. That can be fixed in a later PR though.

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
Yumnah2: enableLimitSwitch() is set to false so it shouldn't impact the motors

bobtseattle (dismissed):
LGTM

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
Yumnah2: Currently there are all of the sensors except the back one

src/main/java/frc/team2412/robot/Controls.java
jamesdooley4: Need to change it back

src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -176,8 +177,10 @@ private void bindLauncherControls() {
 	}
 
 	public void vibrateDriveController(double vibration) {
-		if (!DriverStation.isAutonomous()) {
-			driveController.getHID().setRumble(RumbleType.kBothRumble, vibration);
-		}
+		// if (!DriverStation.isAutonomous()) {
+
+		// }
```
jamesdooley4: Should delete commented-out code

jamesdooley4: Or put it back so the controllers don't rumble in autonomous

jamesdooley4 (dismissed):

bobtseattle (dismissed):
Let's get this in so we can merge with ingest disable code.

jamesdooley4:
@Yumnah2 - I pushed commits to resolve the merge conflicts & autofix the spotless syntax updates

jamesdooley4 (approved):

PR #75, opened 2024-02-28
Vision improvements

KangarooKoala:
Not tested on the robot yet (Need to check accuracy and performance)

### DIFF
--- a/src/main/java/frc/team2412/robot/sensors/AprilTagsProcessor.java
+++ b/src/main/java/frc/team2412/robot/sensors/AprilTagsProcessor.java
@@ -100,6 +100,9 @@ public class AprilTagsProcessor {
 				new PhotonPoseEstimator(
 						fieldLayout, PoseStrategy.MULTI_TAG_PNP_ON_COPROCESSOR, photonCamera, ROBOT_TO_CAM);
 
+		photonPoseEstimator.setLastPose(aprilTagsHelper.getEstimatedPosition());
+		photonPoseEstimator.setMultiTagFallbackStrategy(PoseStrategy.CLOSEST_TO_LAST_POSE);
+
 		networkTables.addListener(
 				networkTables
 						.getTable("photonvision")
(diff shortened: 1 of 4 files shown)


src/main/java/frc/team2412/robot/sensors/AprilTagsProcessor.java
```diff
@@ -49,8 +50,62 @@ public class AprilTagsProcessor {
...
+
+	private static final double MAX_POSE_AMBIGUITY = 0.1;
+
+	// Both of these are in radians
+	// Negate pitch to go from robot/target to cam pitch to cam to robot/target pitch
+	private static final double EXPECTED_CAM_TO_TARGET_PITCH = -ROBOT_TO_CAM.getRotation().getY();
+	private static final double CAM_TO_TARGET_PITCH_TOLERANCE = 0.1;
+
+	private static boolean hasCorrectPitch(Transform3d camToTarget) {
+		return Math.abs(camToTarget.getRotation().getY() - EXPECTED_CAM_TO_TARGET_PITCH)
```
TAKBS2412, line 71: Might be good to log the reported pitch, to debug any errors with tags being incorrectly filtered out/included.

src/main/java/frc/team2412/robot/sensors/AprilTagsProcessor.java
```diff
@@ -136,7 +136,7 @@ public AprilTagsProcessor(DrivebaseWrapper aprilTagsHelper) {
 		photonCamera = new PhotonCamera(Hardware.PHOTON_CAM);
 		photonPoseEstimator =
 				new PhotonPoseEstimator(
-						fieldLayout, PoseStrategy.MULTI_TAG_PNP_ON_COPROCESSOR, photonCamera, ROBOT_TO_CAM);
+						fieldLayout, PoseStrategy.MULTI_TAG_PNP_ON_RIO, photonCamera, ROBOT_TO_CAM);
```
TAKBS2412, line 140: Did you try the coprocessor strategy with the UI option selected? https://github.com/PhotonVision/photonvision/blob/7cec1413410c8053dafd8ecb57095f47469c758c/photon-lib/py/photonlibpy/photonPoseEstimator.py#L36

KangarooKoala: No- The reason I don't think `MULTITAG_PNP_ON_COPROCESSOR` won't work all the time is that since our filters are applied on the Rio side, the result would need to be recalculated on the Rio if the filter changes anything (and thus invalidates the multitag result from the coprocessor).
Ideally we could detect when our filter does anything so that we only calculate on Rio when we have to, but that might take too long to implement for week 1. If the multitag processing takes too long on the Rio, another option is to simply ignore the entire result if anything's bad about it.

jamesdooley4:
@KangarooKoala @jbko6 - I marked this as 'ready for review' as it's what https://github.com/robototes/Crescendo2024/pull/76 is built off of and what was being run today at the Steamnasium. Up to you if you want to merge this in first, and then other - or abandon this one for the larger Steamnasium PR with all the changes in it.

KangarooKoala:
Unless we squash merge both of them, I don't see any advantage to merging them in separately.

PR #89, opened 2024-03-12
Auto field improvements

### LINKED ISSUE #86: Selected auto path keeps running after match starts
@jbko6  - looking at the wpilog logs from E9 match at Glacier Peak, I noticed the "NT\Shuffleboard\Match\Selected auto\Robot" values keep cycling even after the match started. Can you see how to stop those once the robot is enabled, to save some CPU time and memory on the Rio?

KangarooKoala:
Fixes #86
Improves handling of starting rotation (Uses preview starting rotation as a fallback)
Adds epoch logging to the auto field
Adds support for named commands
Adds support for autos with no paths (but with a specific starting pose)


TAKBS2412 (dismissed):
Might not be for this PR, but the LowSpeakerCenterLineN5N4 auto mode doesn't update the displayed robot pose and prints the following in simulation:

```
Warning at com.pathplanner.lib.auto.NamedCommands.getCommand(NamedCommands.java:65): PathPlanner attempted to create a command 'LowSpeakerCenterLineN5N4' that has not been registered with NamedCommands.registerCommand
```

kirbt (approved):

bobtseattle (approved):
LGTM

PR #91, opened 2024-03-13
PIDFF improvements for pivot (IN TESTING)

iamawesomecat:


src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -134,8 +144,11 @@ public void configMotors() {
 		// PID
 		launcherAngleOnePIDController.setP(5.0);
 		launcherAngleOnePIDController.setI(0);
-		launcherAngleOnePIDController.setD(0);
-		launcherAngleOnePIDController.setFF(0);
+		launcherAngleOnePIDController.setD(3);
+		launcherAngleOnePIDController.setFF(
```
jbko6: FF should be recalculated every time the angle of the pivot is set. You can include FF as a parameter when you set the motor reference to accomplish that

iamawesomecat: Okay, so would i just run everything in .setFF in .setReference?

jbko6: Pretty much. See https://github.com/robototes/Crescendo2024/pull/92/files#diff-d8d7d840be58e4901f0702d7c23a5a27cf7cbb95f4b21d2ce2fa8e4ab36a76e9R189 for an example

iamawesomecat: image

would this work?

jbko6: I think so 👍

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -243,7 +244,13 @@ public void setAngleManual(double joystickInput) {
...
 		if (Units.degreesToRotations(getAngle()) > 0.71
 				&& Units.degreesToRotations(getAngle()) < 0.95) {
-			launcherAngleOnePIDController.setReference(manualAngleSetpoint, ControlType.kPosition);
+			launcherAngleOnePIDController.setReference(
+					manualAngleSetpoint,
+					ControlType.kPosition,
+					0,
+					launcherAngleFF.calculate(
+							Units.rotationsToRadians(launcherAngleEncoder.getPosition() - FF_PIVOT_OFFSET),
+							launcherAngleEncoder.getVelocity()));
```
TAKBS2412: Note that the feedforward is based on the setpoint positiion and velocity, so we should be passing in the desired velocity instead of the current velocity, so this should be 0 (same for the other occurrence).

Would also recommend refactoring so the `setReference()` and other logic isn't duplicated.

TAKBS2412 (changes requested):

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: (i know its 15 to midnight BUT does this work?)
image

TAKBS2412: Off the top of my head, yes. But best to make the change and push then I can double check.

src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -59,8 +62,9 @@ public Controls(Subsystems s) {
 
 		launcherAmpPresetButton = codriveController.x();
 		launcherSubwooferPresetButton = codriveController.a();
+		launcherLowerPresetButton = codriveController.y();
 		// launcherPodiumPresetButton = codriveController.povLeft();
-		// launcherTrapPresetButton = codriveController.y();
+		launcherTrapPresetButton = codriveController.y();
```
TAKBS2412: Note that this and the `launcherLowerPresetButton` are both bound to the same button.

src/main/java/frc/team2412/robot/commands/launcher/SetPivotCommand.java
```diff
@@ -0,0 +1,26 @@
...
+
+	public SetPivotCommand(LauncherSubsystem launcherSubsystem, double angle) {
+		this.launcherSubsystem = launcherSubsystem;
+		this.angle = angle;
+		addRequirements(launcherSubsystem);
+	}
+
+	@Override
+	public void initialize() {
+		launcherSubsystem.setAngle(angle);
```
TAKBS2412, line 19: This command doesn't change the manual setpoint, so when it ends the pivot will go to the manual setpoint instead of holding the given position.

It may be better to have only one setpoint variable instead of having two separate variables for the manual and automated angle setpoints to avoid bugs like this.

src/main/java/frc/team2412/robot/util/TrapAlign.java
```diff
@@ -30,20 +32,20 @@ public class TrapAlign {
...
 		Pose2d robotPose = drivebaseSubsystem.getPose();
-		Pose2d trapPose =
-				robotPose.nearest(
-						List.of(
-								(DriverStation.getAlliance().get().equals(Alliance.Blue))
-										? BLUE_TRAP_POSES
-										: RED_TRAP_POSES));
-
+		boolean isBlue;
+		isBlue = DriverStation.getAlliance().get().equals(Alliance.Blue);
```
TAKBS2412: We should add a check for if the optional is empty, otherwise calling `get()` on an empty optional will throw an exception:

```
Error at java.base/java.util.Optional.get(Optional.java:143): Unhandled exception: java.util.NoSuchElementException: No value present
        at java.base/java.util.Optional.get(Optional.java:143)
        at frc.team2412.robot.util.TrapAlign.trapAlign(TrapAlign.java:43)
        at frc.team2412.robot.util.TrapAlign$AlignCommand.initialize(TrapAlign.java:85)
        at edu.wpi.first.wpilibj2.command.CommandScheduler.initCommand(CommandScheduler.java:166)
        at edu.wpi.first.wpilibj2.command.CommandScheduler.schedule(CommandScheduler.java:223)
        at edu.wpi.first.wpilibj2.command.CommandScheduler.schedule(CommandScheduler.java:234)
        at edu.wpi.first.wpilibj2.command.Command.schedule(Command.java:460)
        at edu.wpi.first.wpilibj2.command.button.Trigger$1.run(Trigger.java:69)
        at java.base/java.lang.Iterable.forEach(Iterable.java:75)
        at edu.wpi.first.wpilibj.event.EventLoop.poll(EventLoop.java:37)
        at edu.wpi.first.wpilibj2.command.CommandScheduler.run(CommandScheduler.java:271)
        at frc.team2412.robot.Robot.robotPeriodic(Robot.java:136)
        at edu.wpi.first.wpilibj.IterativeRobotBase.loopFunc(IterativeRobotBase.java:381)
        at edu.wpi.first.wpilibj.TimedRobot.startCompetition(TimedRobot.java:131)
        at edu.wpi.first.wpilibj.RobotBase.runRobot(RobotBase.java:365)
        at edu.wpi.first.wpilibj.RobotBase.lambda$startRobot$0(RobotBase.java:433)
        at java.base/java.lang.Thread.run(Thread.java:833)
```

Arguably, this is unlikely to happen in a match, since alliance color is typically set by the time autonomous starts, but probably best to be safe.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
TAKBS2412: Looking at this again, we should also update this to use the position setpoint instead of the current position from the encoder (this is also mentioned in the docs [here](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/math/controller/ArmFeedforward.html#calculate(double,double))).

Interestingly, last year's version also had this issue (for the wrist, we didn't use `ArmFeedforward` for the arm) - it's possible we were calling `setReference()` often enough that this didn't matter (since the feedforward would've been constantly updated), or that the motors were strong enough to control the (fairly light) wrist without much trouble. Probably best not to rely on either of those things being true this year though.

TAKBS2412 (changes requested):

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
iamawesomecat: image

would this work?

src/main/java/frc/team2412/robot/Controls.java
iamawesomecat: Temporary binding for testing, ill talk to drive team to figure out what the actual binding will be if/when it actually works

src/main/java/frc/team2412/robot/util/TrapAlign.java
iamawesomecat: image
 this good?

TAKBS2412 (commented):
LGTM pending merge conflicts.

src/main/java/frc/team2412/robot/Controls.java
iamawesomecat: fixed

TAKBS2412 (approved):

PR #95, opened 2024-03-19
Vision improvements

KangarooKoala:
Changes:
* Update to PhotonVision 2024.2.9 (This was done to the Orange Pi as well)
* Incorporate rotation into `ROBOT_TO_CAM` instead of rotating the camera stream (Rotating the camera stream messed with calibration)
* Fix pitch check and make it check roll as well
* Separate raw and filtered results and improve look of pose display

Notes:
* 31a7fa0 was tested on Crane but not on Toucan as of 3/18
* We may need to change the roll later depending on how we mount the camera on Toucan


TAKBS2412 (approved):
Some (non-blocking) logging improvements we may want to consider in the future:
- Convert the 3d pose on field entry to a format that AdvantageScope can understand in its 3d view (to make debugging roll/pitch issues easier)
- Log the individual filter status, so we have more fine-grained information on why a particular target was included (a bit trickier since that implies many different entries)

PR #97, opened 2024-03-23
Pre sammy testing

kirbt:


okay
KangarooKoala, line 1: What's the point of this file? If we want to quickly merge this into main then it's fine to leave it in, but if not we probably should delete this. (Which seems to be terminal output from `git branch`, based on the ^[ character and ANSI escape sequences in there)

src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -176,7 +176,10 @@ private void bindLauncherControls() {
 												* LauncherSubsystem.ANGLE_MAX_SPEED));
 
 		launcherLowerPresetButton.onTrue(
-				new SetPivotCommand(s.launcherSubsystem, LauncherSubsystem.RETRACTED_ANGLE));
+				s.launcherSubsystem
+						.run(s.launcherSubsystem::stopLauncher)
+						.until(() -> true)
```
KangarooKoala: (Nit- Ignore if we want to merge this in quickly)
```suggestion
						.runOnce(s.launcherSubsystem::stopLauncher)
```

src/main/java/frc/team2412/robot/util/auto/AutoLogic.java
```diff
@@ -60,8 +60,11 @@ public static enum StartPosition {
 		}
 	};
 
+	// TODO: might be a duplicate, keep until after comp
```
KangarooKoala, line 63: Just a note for later: Even if `AutoLogic.registerCommands()` is the first use of anything in `AutoLogic` (so that the call is when the class is loaded), the commands are initialized before the method is called, so the static initialization block is probably the right way to go. Maybe have a separate method named `init()` ([like we did in 2022](https://github.com/robototes/RapidReact2022/blob/main/src/main/java/frc/team2412/robot/commands/autonomous/JackFiveBallAutoCommand.java#L104)) which will just forced the class to be loaded? We could put `registerCommands()` in the nested class so that calling it won't load the static path lists, but that seems janky since we do always want the register commands before we load the static paths.
**Regardless, I do agree with the sentiment that this should wait until after comp.**

src/main/java/frc/team2412/robot/util/auto/AutoLogic.java
```diff
@@ -300,9 +300,7 @@ public static boolean chooserHasAutoSelected() {
 	}
 
 	public static Command getSelectedAuto() {
-		return Commands.sequence(
-				Commands.waitSeconds(autoDelayEntry.getDouble(0)),
-				availableAutos.getSelected().getAutoCommand());
+		return availableAutos.getSelected().getAutoCommand();
```
KangarooKoala, line 305: Does drive team know that the delay doesn't do anything now? (And is it worth it to update the time estimate?)

src/main/java/frc/team2412/robot/commands/launcher/FullTargetCommand.java
```diff
@@ -32,6 +33,7 @@ public class FullTargetCommand extends Command {
 	private Command yawAlignmentCommand;
 	private Rotation2d yawTarget;
 	private Controls controls;
+	static GenericEntry distanceEntry;
```
KangarooKoala: Nit: This doesn't seem to be used and should be deleted.

KangarooKoala (commented):
The only thing worth thinking about before/during comp is the AutoLogic comment about ignoring the auto delay, the other stuff can wait until after comp.

src/main/java/frc/team2412/robot/util/auto/AutoLogic.java
kirbt: will comment it out until fixed

src/main/java/frc/team2412/robot/Controls.java
kirbt: agree wit this

src/main/java/frc/team2412/robot/sensors/AprilTagsProcessor.java
```diff
@@ -72,39 +63,16 @@ private static boolean hasValidRotation(Transform3d camToTarget) {
...
-				target.getMinAreaRectCorners(),
-				target.getDetectedCorners());
-	}
-
-	private static PhotonPipelineResult filteredPipelineResult(PhotonPipelineResult result) {
-		var copy = copy(result);
-		for (int i = copy.targets.size() - 1; i >= 0; --i) {
-			var target = copy.targets.get(i);
+	private static boolean resultIsValid(PhotonPipelineResult result) {
+		for (var target : result.targets) {
```
TAKBS2412, line 67: @KangarooKoala I realized that we could probably use Java's [Stream API](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html) for this, but that's super unnecessary for competition (this is a random 1 am thought, just advertising as "this is something cool we could do later in the future post Sammamish").

TAKBS2412 (commented):
LGTM, but we do need to run spotless...

src/main/java/frc/team2412/robot/sensors/AprilTagsProcessor.java
KangarooKoala: I've heard the stream API is pretty inefficient (though this one wouldn't need that many intermediate objects so maybe it wouldn't be as bad), but yeah we could look into that later.

TAKBS2412 (approved):

PR #98, opened 2024-03-24
Sammy auto changes

kirbt:
**DONT MERGE TILL EVERYTHING IS TESTED !!! (or we could wait until after comp idk)**

- Made it so that instead of setting launcher angle at subwoofer, will prepare the launcher and flywheel beforehand, allowing for potential time efficiency in auto. 

- Named commands were changed slightly

- Commenting for TODOs to figure out in auto logic

- positioned vision launch test to be real(initiL pose now located on Autoline N2) and testable/debuggable

again, i think these changes should probably be tested before merging into the competition branch incase they ruin things for fun.


KangarooKoala (changes requested):
Requesting changes just so we don't accidentally merge.

Oh huh, guess that didn't work because of perms.

PR #101, opened 2024-03-29
Full target tuning

KangarooKoala:
I don't know what the plan is for merging in the work from 3/28, but here's a PR with the full target changes.

Marking as draft since there's some changes that drastically change (break from a comp POV) controls to make tuning easier.

PR #102, opened 2024-03-29
Vision fixes

KangarooKoala:
Included in #101, just making a separate PR in case the other one takes a while.


TAKBS2412 (dismissed):
LGTM - though we should probably double check this doesn't affect driving in any way, just to be sure.

PR #105, opened 2024-03-30
auto logic adjustments

kirbt:
checker for when toucan ready to launch


src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -108,6 +111,13 @@ public Controls(Subsystems s) {
 			driveController
 					.leftBumper()
 					.whileTrue(new FullTargetCommand(s.launcherSubsystem, s.drivebaseSubsystem, this));
+			driveController
+					.leftBumper()
+					.onTrue(
+							new FullTargetCommand(s.launcherSubsystem, s.drivebaseSubsystem, this)
+									.until(AutoLogic.isReadyToLaunch())
+									.andThen(new WaitCommand(AutoLogic.FEEDER_DELAY))
+									.andThen(new FeederInCommand(s.intakeSubsystem).until(AutoLogic.untilNoNote())));
```
KangarooKoala: Why was this change made? (Note that when the left bumper is pressed, the above binding will be run, scheduling the `FullTargetCommand`, and then this binding will be run, scheduling the `FullTargetCommand.until(...).andThen(...)` composition and canceling the `FullTargetCommand`. ([`Trigger`s](https://github.com/wpilibsuite/allwpilib/blob/main/wpilibNewCommands/src/main/java/edu/wpi/first/wpilibj2/command/button/Trigger.java) are polled using [`EventLoop`s](https://github.com/wpilibsuite/allwpilib/blob/main/wpilibj/src/main/java/edu/wpi/first/wpilibj/event/EventLoop.java), which use a [`LinkedHashSet`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/LinkedHashSet.html), which guarantees iteration order matches insertion order, or the order the bindings were made))

src/main/java/frc/team2412/robot/subsystems/DrivebaseSubsystem.java
```diff
@@ -334,7 +334,7 @@ private void initShuffleboard() {
 						.getEntry();
 		xWheelsEnabled = xWheelsEntry.getBoolean(true);
 		flipTranslationEntry =
-				drivebaseTab
+				Shuffleboard.getTab("Match")
```
KangarooKoala: Is this something that needs to be on the Match tab? On the one hand, we do want DT to be able to use it quickly when they have to, but they also shouldn't have to use it.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -357,6 +358,11 @@ private void initShuffleboard() {
 
 		angleSetpointEntry =
 				Shuffleboard.getTab("Launcher").add("Angle Setpoint", 0).withPosition(2, 2).getEntry();
+
+		Shuffleboard.getTab("Launcher")
+				.addDouble("setpoint + tol", () -> angleSetpoint - RPM_TOLERANCE);
```
KangarooKoala: The wording doesn't match the code? (+ vs -)
Also, this should have a position (and size) specified to make sure this doesn't overlap with other entries.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -357,6 +358,11 @@ private void initShuffleboard() {
...
+
+		Shuffleboard.getTab("Launcher").addBoolean("yay", this::isAtSpeed);
```
KangarooKoala: If this is supposed to stay in the code, please give this a more descriptive name and specify a position (and size) so that it doesn't overlap with other entries.

src/main/java/frc/team2412/robot/util/auto/AutoLogic.java
```diff
@@ -325,6 +293,14 @@ public static boolean chooserHasAutoSelected() {
 	}
 
 	public static Command getSelectedAuto() {
+
+		// TODO: figure out why code crashes when this sequential command group is returned
+		// code also crashes when i run without wait command, something about composition error? idk
```
KangarooKoala: It isn't allowed to use the same command in multiple compositions. Generally, this is avoided by recreating commands each time, which in this case probably looks like making `AutoPath` create the command in `getAutoCommand()` instead of making it once in the constructor and reusing it.

src/main/java/frc/team2412/robot/subsystems/DrivebaseSubsystem.java
kirbt: probably dont need it anymore since the translation issue durring sammy is fixed, i guess ill put it in drivebase for now

src/main/java/frc/team2412/robot/Controls.java
kirbt: this was used for debugging auto, ill uncomment it until we need to debug that again

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
kirbt: removed cuase i dont need it anymore

src/main/java/frc/team2412/robot/util/auto/AutoLogic.java
kirbt: It would be useful for drive team to be able to set a delay timer before an auto starts without programming having to add the delay via hardcode/pathplanner. How would you suggest getting around said error?

KangarooKoala: What I just said- Change `AutoPath.getAutoCommand()` to `return AutoBuilder.buildAuto(pathPlannerAutoName)` and remove `autoCommand`.

src/main/java/frc/team2412/robot/util/auto/AutoLogic.java
```diff
@@ -356,4 +332,176 @@ public static double getEstimatedAutoDuration() {
...
+								Commands.either(
+										new SetAngleLaunchCommand(
+														s.launcherSubsystem,
+														LauncherSubsystem.SPEAKER_SHOOT_SPEED_RPM,
+														LauncherSubsystem.SUBWOOFER_AIM_ANGLE)
+												.until(isReadyToLaunch())
+												.andThen(new WaitCommand(FEEDER_DELAY))
+												.andThen(new FeederInCommand(s.intakeSubsystem).until(untilNoNote()))
+												.andThen(new WaitCommand(0.4))
+												.withName("Auto - SubwooferLaunchCommand"),
```
TAKBS2412: For these and the other command compositions, we should call `withName()` on the outermost command, and not a command inside the composition, as the outermost command will be the one that is logged.

bobtseattle (commented):
Now that we have some logged launcher data points, it would be good to make sure we are launching from those poses.
We should also limit the launcher angle changes, unless they are necessary to go under the stage or shoot from a different position.  The robot has been picking up fine with the launcher at subwoofer angle.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -371,6 +372,7 @@ public void periodic() {
 		launcherIsAtSpeed.setBoolean(isAtSpeed());
 		launcherAngleManual.setDouble(manualAngleSetpoint);
 		angleSetpointEntry.setDouble(angleSetpoint);
+		setLauncherSpeedEntry.setDouble(rpmSetpoint);
```
KangarooKoala: Have we tested this? I'm not sure about the back-and-forth here between sets from Shuffleboard and sets from robot code.

TAKBS2412: Also note that #110 also logs the setpoint, so the two PRs will likely conflict with each other.

kirbt: true, ill remove said changes later to avoid conflicts

TAKBS2412 (dismissed):

TAKBS2412 (approved):

PR #110, opened 2024-03-30
Launcher setpoint logging

iamawesomecat:


TAKBS2412 (dismissed):

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -357,6 +359,8 @@ private void initShuffleboard() {
 
 		angleSetpointEntry =
 				Shuffleboard.getTab("Launcher").add("Angle Setpoint", 0).withPosition(2, 2).getEntry();
+		launcherFlywheelSetpointEntry =
+				Shuffleboard.getTab("Launcher").add("Flywheel Setpoint", 0).getEntry();
```
KangarooKoala: Should we specify the position to make sure this doesn't overlap with anything?

bobtseattle (commented):
Seems like it would be good to log both flywheels.
Though, the setpoint will be the same for both unless using the amp, so nevermind.

TAKBS2412 (approved):

PR #111, opened 2024-04-02
Sysid Tuning

jbko6:
I threw out the angle sysid tuning as it seemed to be overshooting and we was running out of time.


bobtseattle (approved):
LGTM

src/main/java/frc/team2412/robot/subsystems/DrivebaseSubsystem.java
```diff
@@ -48,7 +48,7 @@ public class DrivebaseSubsystem extends SubsystemBase {
 	public static final double MAX_SPEED =
 			Robot.getInstance().getRobotType() == RobotType.BONK
 					? 3.0
-					: Robot.getInstance().getRobotType() == RobotType.PRACTICE
+					: Robot.getInstance().getRobotType() == RobotType.COMPETITION
 							? 6.0
```
jamesdooley4, line 52: @jbko6 - I realized that all the kV values depend on this constant (since kV * MAX_SPEED should be about 12.0V), which likely should also have been calculated during sysId testing (it's simply the speed the robot went when at full power).

TAKBS2412 (approved):
LGTM. In an ideal world we would tune this further, but it was driveable when we tested it during the meeting.

src/main/java/frc/team2412/robot/subsystems/DrivebaseSubsystem.java
jbko6: That would explain why it was giving more kV than it should have. If we have a chance at DCMPs with nothing better to do I'll try reverting to the originally calculated kV value with the correct max speed to see how it performs.

jamesdooley4: @jbko6 - out of curiosity, what is the current physical max speed of the robot?

PR #114, opened 2024-04-03
Changes from 4/2 testing

jamesdooley4:


src/main/deploy/deploy
```diff
@@ -1,3 +1,2 @@
-Files placed in this directory will be deployed to the RoboRIO into the
```
jamesdooley4: ? Mistaken deleted line?

KangarooKoala: Probably? It was done in 66d7809 (pivot offsets), which wouldn't have a reason to delete that line (especially since it left the other lines in the file).

jamesdooley4:
LGTM. I can't approve since I created the PR, so waiting for an approval before merging.

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
```diff
@@ -243,6 +285,28 @@ public boolean intakeRightSeesNote() {
...
+		// 	return true;
+		// }
+		return intakeFrontSensorDebouncer.calculate(intakeFrontSensor.isPressed());
+	}
+
+	public boolean debouncedIntakeLeftSensor() {
+		// if (intakeLeftSensorDebouncer.calculate(intakeLeftSensor.isPressed())) {
+		// 	return true;
+		// }
+		return intakeLeftSensorDebouncer.calculate(intakeFrontSensor.isPressed());
```
TAKBS2412, line 300: For after DCMP (since we don't have the sensors on the robot right now): this should be checking `intakeLeftSensor.isPressed()` (same for the right sensor method below).

TAKBS2412 (approved):

PR #115, opened 2024-04-08
DCMP Changes

jbko6:


src/main/java/frc/team2412/robot/commands/launcher/SyncPivotRelativeEncoderCommand.java
```diff
@@ -0,0 +1,56 @@
+package frc.team2412.robot.commands.launcher;
+
+import edu.wpi.first.math.filter.MedianFilter;
+import edu.wpi.first.wpilibj2.command.Command;
+import frc.team2412.robot.subsystems.LauncherSubsystem;
+
+public class SyncPivotRelativeEncoderCommand extends Command {
+
+	private int MAX_SAMPLES = 50; // 6 seconds, 50 per second
```
KangarooKoala: The duration in this comment doesn't make sense? It's correct that there should be 50 samples per second, but that means 50 samples is 1 second, not 6.

src/main/java/frc/team2412/robot/commands/launcher/SyncPivotRelativeEncoderCommand.java
```diff
@@ -0,0 +1,56 @@
...
+	private final double ANGLE_SPEED_TOLERANCE = 0.1;
+	private final MedianFilter filter;
+
+	private final LauncherSubsystem launcherSubsystem;
+
+	private int samples;
+
+	public SyncPivotRelativeEncoderCommand(
+			LauncherSubsystem launcherSubsystem, boolean ignorePrevOffset) {
```
KangarooKoala: Is `ignorePrevOffset` supposed to be used for anything?

src/main/java/frc/team2412/robot/Robot.java
```diff
@@ -121,6 +122,12 @@ public void robotInit() {
 		dashboard = new MatchDashboard(subsystems);
 
 		RobotController.setBrownoutVoltage(5.75);
+
+		CommandScheduler.getInstance()
+				.schedule(
+						new SyncPivotRelativeEncoderCommand(subsystems.launcherSubsystem)
+								.ignoringDisable(true)
+								.withName("Sync Relative Encoder"));
```
KangarooKoala, line 131: This is only meant to be done once shortly after robot code startup, right?

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
KangarooKoala, line 1: When we have a bit of a breather, it'd be good to put in `FRONT_MOTOR_ENABLED`, `FRONT_SENSOR_ENABLED`, etc. constants to make it easier to enable and disable intakes as needed. Might be a thing to do after we ship the robot, though (since we could have high enough confidence in it working to need less testing, plus it's not as high priority as other items planned for today and tomorrow).

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -278,7 +293,22 @@ public double getAngleSpeed() {
...
+		this.ignoreLimits = ignoreLimits;
+		if (powerControl || ignoreLimits) {
+			launcherAngleOneMotor.set(ignoreLimits ? joystickInput * 0.3 : joystickInput);
+			manualAngleSetpoint =
+					MathUtil.clamp(
+							Units.degreesToRotations(getAngle()),
+							PIVOT_SOFTSTOP_BACKWARD,
+							PIVOT_SOFTSTOP_FORWARD);
+			return;
+		}
```
KangarooKoala: Why we're setting `manualAngleSetpoint` when we're using power control?

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -316,6 +346,22 @@ private void initShuffleboard() {
...
+		Shuffleboard.getTab("Launcher")
+				.addDouble(
+						"relative encoder", () -> launcherAngleOneMotor.getEncoder().getPosition() * 360);
+		Shuffleboard.getTab("Launcher")
+				.addDouble("relative encoder offset", () -> relativeEncoderStartPosition.orElse(0.0));
+
+		Shuffleboard.getTab("Launcher.add")
+				.addDouble(
+						"relative encoder",
+						() -> Units.rotationsToDegrees(launcherAngleOneMotor.getEncoder().getPosition()));
```
KangarooKoala: Why is this in a differently named tab? Is it redundant with the "relative encoder" entry?

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -394,12 +440,31 @@ private void initShuffleboard() {
...
+	}
+
+	public void zeroRelativeEncoder(double pivotAngle) {
+
+		double currentRelativePosition =
+				Units.rotationsToDegrees(launcherAngleOneMotor.getEncoder().getPosition());
+		double offset = pivotAngle - currentRelativePosition;
+
+		if (relativeEncoderStartPosition.isEmpty()
+				|| Math.abs(relativeEncoderStartPosition.orElse(0.0) + offset) > OFFSET_SYNCING_TOLERANCE) {
```
KangarooKoala: Do we want to add or subtract?

src/main/deploy/swerve/modules/physicalproperties.json
```diff
@@ -2,7 +2,7 @@
   "optimalVoltage": 12,
   "wheelGripCoefficientOfFriction": 1.19,
   "currentLimit": {
-    "drive": 60,
+    "drive": 140,
```
jamesdooley4: Should drop this to 120, as was consistently reaching brownout conditions near the end of a match. Can do after the PR.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -410,13 +475,35 @@ public void periodic() {
 		angleSetpointEntry.setDouble(angleSetpoint);
 		launcherFlywheelSetpointEntry.setDouble(rpmSetpoint);
 
-		// sanity check the pivot encoder
+		// PIVOT ENCODER SANITY CHECKS
+		// compares the relative encoder angle vs the absolute encoder angle
+		if (relativeEncoderStartPosition.isPresent()
+				&& Math.abs(getAngle() - getAngleOneMotorAngle()) <= ENCODER_DIFFERENCE_TOLERANCE) {
```
KangarooKoala: `<=` or `>=`?

src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -250,7 +251,7 @@ private void bindLauncherControls() {
 		// 				s.launcherSubsystem.runEnd(
 		// 						s.launcherSubsystem::launch, s.launcherSubsystem::stopLauncher));
 
-		driveController.b().onTrue(new InstantCommand(() -> s.launcherSubsystem.launch(4000)));
+		driveController.b().onTrue(new InstantCommand(() -> s.launcherSubsystem.launch(6500)));
```
jamesdooley4, line 247: Note that driveController.b is mapped to both this & the driveIntakeStopButton. Doing both those actions on 'b' seems reasonable, but would be nice to only map the button in one place.

jamesdooley4: Oh, and the bump to 6500 here is specifically for use when launching notes cross-field.

src/main/java/frc/team2412/robot/util/auto/AutoLogic.java
```diff
@@ -474,7 +479,7 @@ public static Command feederIn() {
 		return (INTAKE_ENABLED && LAUNCHER_ENABLED
 						? Commands.waitUntil(isReadyToLaunch())
 								.andThen(Commands.waitSeconds(FEEDER_DELAY))
-								.andThen(new FeederInCommand(s.intakeSubsystem))
+								.andThen(new FeederInCommand(s.intakeSubsystem).until(untilNoNote())
```
KangarooKoala:
```suggestion
								.andThen(new FeederInCommand(s.intakeSubsystem).until(untilNoNote()))
```
Missing closing parentheses is causing the compile error.

src/main/java/frc/team2412/robot/Hardware.java
```diff
@@ -22,17 +22,17 @@ public class Hardware {
...
 
 	public static final int INGEST_MOTOR = 44;
 
 	public static final int INDEX_MOTOR_UPPER = 45;
 
 	public static final int FEEDER_MOTOR = 46;
-	// LED strip is PWM port 8
-	public static final int BLINKIN_LED = 8;
+	// LED strip is PWM port 2
+	public static final int BLINKIN_LED = 2;
```
jamesdooley4: I'm not sure that port 2 works...

src/main/java/frc/team2412/robot/commands/launcher/SyncPivotRelativeEncoderCommand.java
jamesdooley4: Should update comment to match code (1 second)

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -278,7 +293,22 @@ public double getAngleSpeed() {
...
 
-	public void setAngleManual(double joystickInput) {
+	public void restoreLimits() {
+		this.ignoreLimits = false;
+	}
+
+	public void setAngleManual(double joystickInput, boolean powerControl, boolean ignoreLimits) {
+		this.ignoreLimits = ignoreLimits;
+		if (powerControl || ignoreLimits) {
+			launcherAngleOneMotor.set(ignoreLimits ? joystickInput * 0.3 : joystickInput);
```
jamesdooley4: Recommend bumping the multiplier to 0.5, so the motor has enough power to overcome a bent bellypan

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
jamesdooley4: The `<=` seems wrong, as that would indicate that things are in sync (as the difference is small).

src/main/java/frc/team2412/robot/util/auto/AutoLogic.java
```diff
@@ -474,7 +479,7 @@ public static Command feederIn() {
...
 								.andThen(Commands.waitSeconds(0.1))
```
jamesdooley4: Would be nice to name this "feed until note launched", to differentiate from "feed until note detected". I assumed `feederIn` meant the latter. And bonus if you add a `.withName("myFancyName")` so the command has a meaningful name in the logs.

src/main/java/swervelib/motors/TalonFXSwerve.java
```diff
@@ -408,7 +408,9 @@ public void setCurrentLimit(int currentLimit) {
 		cfg.refresh(configuration.CurrentLimits);
 		cfg.apply(
 				configuration.CurrentLimits.withStatorCurrentLimit(currentLimit)
-						.withStatorCurrentLimitEnable(true));
+						.withStatorCurrentLimitEnable(true)
+						.withSupplyCurrentLimit(80)
+						.withSupplyCurrentLimitEnable(true));
```
jamesdooley4: @jbko6 - this really ought to be configurable so it can be different between the drive & steer motors.

src/main/java/frc/team2412/robot/util/auto/AutoLogic.java
jamesdooley4: Missing closing parenthesis, causing build error.

jamesdooley4 (dismissed):

src/main/java/frc/team2412/robot/commands/launcher/SyncPivotRelativeEncoderCommand.java
KangarooKoala: (Duplicate of James's comment)

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
KangarooKoala: (Duplicate of James's comment)

src/main/java/frc/team2412/robot/Robot.java
kirbt: yes

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
kirbt: because we want the launcher to not snap back to the previous set point after disabling power control

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
kirbt: fumble

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
KangarooKoala: Have separate motor constants, though sensor constants aren't in yet (but also aren't important).

src/main/java/frc/team2412/robot/subsystems/IntakeSubsystem.java
```diff
@@ -170,9 +201,15 @@ public void intakeReverse() {
...
+			}
+			if (RIGHT_MOTOR_ENABLED) {
+				intakeMotorRight.set(INTAKE_IN_SPEED);
+			}
+			if (FRONT_MOTOR_ENABLED) {
+				intakeMotorFront.set(INTAKE_REJECT_SPEED);
+			}
 		}
 		ingestMotor.set(INTAKE_REJECT_SPEED);
 		indexMotorUpper.set(INTAKE_IN_SPEED);
```
jamesdooley4, line 219: I expected the index to also be set to 'reject' speed, so it's spinning the same direction as the center/ingest roller. As written here, they are running in opposite directions, so when each roller grabs the opposite ends of the note they'll try to stretch it.

src/main/java/swervelib/motors/SparkFlexSwerve.java
```diff
@@ -110,11 +110,12 @@ public void setVoltageCompensation(double nominalVoltage) {
...
 	 *
-	 * @param currentLimit Current limit in AMPS at free speed.
+	 * @param statorLimit Current limit in AMPS at free speed.
+	 * @param supplyLimit
 	 */
 	@Override
-	public void setCurrentLimit(int currentLimit) {
-		configureSparkFlex(() -> motor.setSmartCurrentLimit(currentLimit));
+	public void setCurrentLimit(int statorLimit, int supplyLimit) {
+		configureSparkFlex(() -> motor.setSmartCurrentLimit(statorLimit));
```
jamesdooley4: I believe for REV motors, the current limit here is the supply limit

src/main/java/swervelib/motors/SparkMaxBrushedMotorSwerve.java
```diff
@@ -157,11 +157,12 @@ public void setVoltageCompensation(double nominalVoltage) {
...
 	 *
-	 * @param currentLimit Current limit in AMPS at free speed.
+	 * @param statorLimit Current limit in AMPS at free speed.
+	 * @param supplyLimit
 	 */
 	@Override
-	public void setCurrentLimit(int currentLimit) {
-		configureSparkMax(() -> motor.setSmartCurrentLimit(currentLimit));
+	public void setCurrentLimit(int statorLimit, int supplyLimit) {
+		configureSparkMax(() -> motor.setSmartCurrentLimit(statorLimit));
```
jamesdooley4: I believe for REV motors, the current limit here is the supply limit

src/main/java/swervelib/motors/SparkMaxSwerve.java
```diff
@@ -91,11 +91,12 @@ public void setVoltageCompensation(double nominalVoltage) {
...
 	 *
-	 * @param currentLimit Current limit in AMPS at free speed.
+	 * @param statorLimit Current limit in AMPS at free speed.
+	 * @param supplyLimit
 	 */
 	@Override
-	public void setCurrentLimit(int currentLimit) {
-		configureSparkMax(() -> motor.setSmartCurrentLimit(currentLimit));
+	public void setCurrentLimit(int statorLimit, int supplyLimit) {
+		configureSparkMax(() -> motor.setSmartCurrentLimit(statorLimit));
```
jamesdooley4: I believe for REV motors, the current limit here is the supply limit

src/main/java/swervelib/motors/TalonFXSwerve.java
```diff
@@ -400,15 +400,26 @@ public void setVoltageCompensation(double nominalVoltage) {
...
+						.withSupplyCurrentLimit(supplyLimit)
+						.withSupplyCurrentLimitEnable(true));
+		System.out.println(
+				"Configured drivebase motor "
+						+ motor.getDeviceID()
+						+ " with stator limit of "
+						+ statorLimit
+						+ " amps and supply limit of "
+						+ supplyLimit
+						+ " amps.");
```
jamesdooley4: Temp debug code?

src/main/java/swervelib/motors/TalonSRXSwerve.java
```diff
@@ -360,12 +360,13 @@ public void setVoltageCompensation(double nominalVoltage) {
...
+	 * @param statorLimit Current limit in AMPS at free speed.
+	 * @param supplyLimit
 	 */
 	@Override
-	public void setCurrentLimit(int currentLimit) {
-		configuration.continuousCurrentLimit = currentLimit;
-		configuration.peakCurrentLimit = currentLimit;
+	public void setCurrentLimit(int statorLimit, int supplyLimit) {
+		configuration.continuousCurrentLimit = statorLimit;
+		configuration.peakCurrentLimit = statorLimit;
```
jamesdooley4, line 369: FYI, TalonSRX has a `configSupplyCurrentLimit(..)` method that could use the supply value, but I don't think it's worth trying to use it here.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -316,13 +344,18 @@ private void initShuffleboard() {
...
-		launcherIsAtSpeed =
-				Shuffleboard.getTab("Match")
-						.add("flywheels at target speed", false)
-						.withSize(1, 1)
-						.withWidget(BuiltInWidgets.kBooleanBox)
-						.withPosition(0, 2)
-						.getEntry();
+		Shuffleboard.getTab("Launcher")
+				.addDouble(
+						"relative encoder", () -> launcherAngleOneMotor.getEncoder().getPosition() * 360);
```
jamesdooley4, line 422: FYI, looking at logs, this isn't updated when the robot is disabled. I suspect the `.addDouble(..)` way of supplying data only samples the values when the robot is enabled. Nothing needing to change here, just something I noticed from the logs

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -192,6 +202,9 @@ public void configMotors() {
 		launcherAngleTwoMotor.getEncoder().setPosition(launcherAngleEncoder.getPosition());
 		launcherAngleTwoMotor.getEncoder().setPositionConversionFactor(PIVOT_GEARING_RATIO);
```
jamesdooley4, line 232: @kirbt  - these lines (and the ones above) duplicate what the new `zeroRelativeEncoder()` method does. Since the relative encoder appears to support resetting its position, I'd recommend using that instead of keeping track of an offset & manually using that. And we really should look at the robot to see what the PIVOT_GEARING_RATIO value ought to be. It appears to be assuming a 3x3x5 gearbox and a 16 tooth sprocket driving a 64 tooth gear.

jamesdooley4: So, `zeroRelativeEncoder()` could just call `launcherAngleOneMotor.getEncoder().setPosition(...)` with the passed-in position, assuming that there hasn't been meaningful movement between when the reference value was calculated & when the method is called.

TAKBS2412 (commented):
I think having the Amp align changes in this PR and #116 is causing merge conflicts. Would also be worth merging in #117 to this branch.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
kirbt: definitely subtract

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -394,29 +497,83 @@ private void initShuffleboard() {
...
+							+ ". Is overidden: "
+							+ ignoreLimits,
+					false);
+		}
 
-		// sanity check the pivot encoder
-		if (launcherAngleEncoder.getPosition() >= PIVOT_SOFTSTOP_FORWARD + PIVOT_DISABLE_OFFSET
-				|| launcherAngleEncoder.getPosition() <= PIVOT_SOFTSTOP_BACKWARD - PIVOT_DISABLE_OFFSET) {
-			launcherAngleOneMotor.disable();
+		if (launcherAngleEncoder.getPosition()
```
jamesdooley4: Should this be getAngle() instead of always reading the Lamprey encoder?

jamesdooley4 (dismissed):

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
```diff
@@ -557,7 +559,7 @@ public void periodic() {
 					false);
 		}
 
-		if (launcherAngleEncoder.getPosition()
+		if (getPosition()
 						>= (USE_THROUGHBORE ? PIVOT_SOFTSTOP_FORWARD_THROUGHBORE : PIVOT_SOFTSTOP_FORWARD)
 								+ PIVOT_DISABLE_OFFSET
 				|| launcherAngleEncoder.getPosition()
```
jamesdooley4, line 565: Need to update both uses.

jamesdooley4 (approved):

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
jamesdooley4: Resolving, as this isn't critical to fix since it just means the motor's relative encoder has a somewhat random position offset - which is generally the case since it always depends on where the arm is when power is first applied.

src/main/java/frc/team2412/robot/subsystems/LauncherSubsystem.java
jbko6: will do in worlds pr

PR #117, opened 2024-04-11
Cleanup code

KangarooKoala:
Should fix the errorprone warnings (as well as one from VSCode)


KangarooKoala:
TODO: Completely remove FullTargetCommand's `YAW_TARGET_VIBRATION_TOLERANCE` (drive team prefers to go off of visual confirmation for yaw), cleanup `driveJoystick()` to reduce allocations (specifically inversions).
Rest should be good for review though.

KangarooKoala:
Decided against cleaning up `driveJoystick()` since it doesn't seem to actually take up a lot of allocations (based on James's earlier testing), plus it would've affected `DriveToNoteCommand` (maybe for the better, but I'm not sure).

PR #120, opened 2024-06-14
Rotate to apriltag

KangarooKoala:
Ignore the commit message- After testing on Monday 6/10, it was working. Not sure what changed, though, which might be worth investigating.

Main purpose is to have a mode where it rotates towards an AprilTag for use at outreach. The control is done in Shuffleboard to avoid accidental activations. Not terribly well tuned, but it should still be a decently interesting demo.

### DIFF
--- a/src/main/java/frc/team2412/robot/Controls.java
+++ b/src/main/java/frc/team2412/robot/Controls.java
@@ -11,8 +11,8 @@ import static frc.team2412.robot.Subsystems.SubsystemConstants.LED_ENABLED;
 import edu.wpi.first.math.MathUtil;
 import edu.wpi.first.math.geometry.Rotation2d;
 import edu.wpi.first.wpilibj.DriverStation;
-import edu.wpi.first.wpilibj.DriverStation.Alliance;
 import edu.wpi.first.wpilibj.GenericHID.RumbleType;
+import edu.wpi.first.wpilibj.Timer;
 import edu.wpi.first.wpilibj2.command.CommandScheduler;
 import edu.wpi.first.wpilibj2.command.Commands;
 import edu.wpi.first.wpilibj2.command.InstantCommand;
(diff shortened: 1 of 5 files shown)


src/main/java/frc/team2412/robot/Controls.java
```diff
@@ -136,6 +138,25 @@ public Controls(Subsystems s) {
...
+			new Trigger(s.rotateToSpeaker)
+					.whileTrue(
+							s.drivebaseSubsystem
+									.rotateToAngle(
+											() -> {
+												double currentTimestamp = Timer.getFPGATimestamp();
+												double robotToTagAngleTimestamp =
+														s.apriltagsProcessor.getLastRotatedAngleTimestamp();
+												var robotAngle = s.drivebaseSubsystem.getPose().getRotation();
+												var robotToTagAngle = s.apriltagsProcessor.getLastRotatedAngle();
```
markpete, line 151: You might move this after the if() {} below.  No reason to have the computer do the work to get robotToTagAngle if you will never use it when the if is true.

src/main/java/frc/team2412/robot/sensors/AprilTagsProcessor.java
```diff
@@ -164,6 +193,35 @@ public void update() {
...
 
+	private void updateSpeakerAngle(PhotonPipelineResult pipelineResult) {
+		int trackingId = (int) trackingIdEntry.getInteger(-1);
+		for (var target : pipelineResult.targets) {
+			if (target.getFiducialId() != trackingId) {
+				continue;
+			}
+			// Camera is rotated so that global left is camera down
+			lastRotatedAngle = Rotation2d.fromDegrees(-target.getPitch());
+			lastRotatedAngleTimestamp = pipelineResult.getTimestampSeconds();
```
markpete, line 204: Add a break; here.  Once you've found the one you want, there is no reason to continue through the rest of the loop.

PR #122, opened 2024-10-15
Toucan offseason changes

KangarooKoala:
Changes to Toucan code at offseason events


RobototesProgrammers (dismissed):
Test approval

src/main/deploy/swerve/modules/physicalproperties.json
```diff
@@ -11,7 +11,7 @@
   },
   "conversionFactor": {
     "angle": 19.2,
-    "drive": 0.0696793458626792
+    "drive": 0.06658958893604663608562691131498
```
markpete, line 14: Floats only have 7 significant digits (plus an exponent) and doubles have 16 decimal digits (plus an exponent) of precision, so this may be overkill.

https://en.wikipedia.org/wiki/Floating-point_arithmetic#IEEE_754:_floating_point_in_modern_computers

